PR #4, opened 2020-04-02
Update macOS dependencies

RytoEX:
### Description

Update macOS dependencies
* Update Opus from 1.2.1 to 1.3.1
* ~~Update libogg from 1.3.3 to 1.3.4~~
* Update libvpx from 1.7.0 to 1.8.2
* Update Jansson from 2.11 to 2.12
* Update FFmpeg from 4.0.2 to 4.2.2
* Update x264 git repo location

### Motivation and Context

This brings macOS in line with the [current Windows release dependencies](https://github.com/obsproject/obs-studio/wiki/Release-Dependency-Versions/2da5ee806df52234a834499f77e8b722f5adcac5) for Opus, libogg, and FFmpeg.  The Windows deps use libvpx 1.8.1, but 1.8.2 is currently the latest release.  It looks like the obs-studio repo bundles Jansson 2.9, but the macOS CI scripts were already using Jansson 2.11, so I've updated to Jansson 2.12.  I'm open to adjusting libvpx and Jansson versions to something else.

### How Has This Been Tested?

I have only tested that the links for the updated URLs work.  We're going to need to build OBS with the new deps to make sure everything works.

### Types of changes

- CI update

### Checklist:

### DIFF
--- a/build-package-deps-osx.sh
+++ b/build-package-deps-osx.sh
@@ -51,9 +51,9 @@ export MACOSX_DEPLOYMENT_TARGET=10.11
 export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/tmp/obsdeps/lib/pkgconfig
 
 # libopus
-curl -L -O https://ftp.osuosl.org/pub/xiph/releases/opus/opus-1.2.1.tar.gz
-tar -xf opus-1.2.1.tar.gz
-cd ./opus-1.2.1
+curl -L -O https://ftp.osuosl.org/pub/xiph/releases/opus/opus-1.3.1.tar.gz
+tar -xf opus-1.3.1.tar.gz
+cd ./opus-1.3.1
 mkdir build
 cd ./build
 ../configure --disable-shared --enable-static --prefix="/tmp/obsdeps"
@@ -87,10 +87,10 @@ make install
 cd $WORK_DIR
 
 # libvpx
-curl -L -O https://chromium.googlesource.com/webm/libvpx/+archive/v1.7.0.tar.gz
-mkdir -p ./libvpx-v1.7.0
-tar -xf v1.7.0.tar.gz -C $PWD/libvpx-v1.7.0
-cd ./libvpx-v1.7.0
+curl -L -O https://chromium.googlesource.com/webm/libvpx/+archive/v1.8.2.tar.gz
+mkdir -p ./libvpx-v1.8.2
+tar -xf v1.8.2.tar.gz -C $PWD/libvpx-v1.8.2
+cd ./libvpx-v1.8.2
 mkdir -p build
 cd ./build
 ../configure --disable-shared --prefix="/tmp/obsdeps" --libdir="/tmp/obsdeps/lib"
@@ -118,9 +118,9 @@ rsync -avh --include="*/" --include="*.h" --exclude="*" ./* $DEPS_DEST/include/
 cd $WORK_DIR
 
 # janson
-curl -L -O http://www.digip.org/jansson/releases/jansson-2.11.tar.gz
-tar -xf jansson-2.11.tar.gz
-cd jansson-2.11
+curl -L -O http://www.digip.org/jansson/releases/jansson-2.12.tar.gz
+tar -xf jansson-2.12.tar.gz
+cd jansson-2.12
 mkdir build
 cd ./build
 ../configure --libdir="/tmp/obsdeps/bin" --enable-shared --disable-static
@@ -135,9 +135,9 @@ export LDFLAGS="-L/tmp/obsdeps/lib"
 export CFLAGS="-I/tmp/obsdeps/include"
 
 # FFMPEG
-curl -L -O https://github.com/FFmpeg/FFmpeg/archive/n4.0.2.zip
-unzip ./n4.0.2.zip
-cd ./FFmpeg-n4.0.2
+curl -L -O https://github.com/FFmpeg/FFmpeg/archive/n4.2.2.zip
+unzip ./n4.2.2.zip
+cd ./FFmpeg-n4.2.2
 mkdir build
 cd ./build
 ../configure --pkg-config-flags="--static" --extra-ldflags="-mmacosx-version-min=10.11" --enable-shared --disable-static --shlibdir="/tmp/obsdeps/bin" --enable-gpl --disable-doc --enable-libx264 --enable-libopus --enable-libvorbis --enable-libvpx --disable-outdev=sdl
--- a/build-package-deps-osx.sh
+++ b/build-package-deps-osx.sh
@@ -63,10 +63,11 @@ make install
 cd $WORK_DIR
 
 # libogg
-curl -L -O https://ftp.osuosl.org/pub/xiph/releases/ogg/libogg-1.3.3.tar.gz
-tar -xf libogg-1.3.3.tar.gz
-cd ./libogg-1.3.3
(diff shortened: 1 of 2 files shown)


PatTheMav:
Tried to build the deps today, some breakage was found:

* libogg breaks libvorbis.

Libogg 1.3.4 introduced a header-import change that _should_ work on macOS 10.9+, but breaks projects depending on it (e.g. libvorbis). It hasn't been fully discovered why (as building libogg with the missing import works, but building projects including libogg doesn't), but reverting this commit fixes the issue: https://git.xiph.org/?p=ogg.git;a=commitdiff;h=6449883ccacfee276ed9d99fa047342cdc51ab88

(See also: https://trac.macports.org/ticket/58929)

* Bad x264 checkout URL - the repository now sits at code.videolan.org/videolan/x264.git
* Maybe I'm mistaken but there's a `cd $WORK_DIR` missing between FFmpeg and luajit

With those changes the deps build successfully and I will try to build OBS with the changes from https://github.com/obsproject/obs-studio/pull/2617.

dodgepong:
Looks like we need to modify the FFmpeg build flags to enable SRT output.

PatTheMav:
Yeah already on it.. but so far it fails to detect the `srt` library in the temporary `obsdeps` even though `pkgconfig` gives the right static library and version, it's infuriating. But I haven't given up hope yet..

RytoEX:
@PatTheMav 
There is [another commit](https://git.xiph.org/?p=ogg.git;a=commit;h=c8fca6b4a02d695b1ceea39b330d4406001c03ed) recently in libogg that fixes that.  We could just pull the current git instead of the latest tag.

Whoops on x264.  Thought I'd checked that.  That's embarrassing, but I'll fix it.

I hadn't checked the _other_ parts of the scripts, and just assumed they would work.  If there are actual bugs with the parts of the script, then we should address them in other commits/PRs.  As it turns out, Azure no longer supports the macos-10.13 vmImage, and it will silently fail if that is set, so #3 needs to be merged before Azure will actually run anything here.

RytoEX:
I've removed the update to libogg for now, leaving it at 1.3.3.  I'm open to changing it to a git checkout for the current git master, but I'm also open to leaving it at 1.3.3.  Though apparently, the Windows release deps use [a specific commit (df53eebf72)](https://git.xiph.org/?p=ogg.git;a=commit;h=df53eebf72e86eb179465f53dd77297ae72ae233) between 1.3.3. and 1.3.4.

RytoEX:
I'm actually going to split off the x264 git repo update so that Azure won't immediately fail once we fix the pipeline.

RytoEX:
Split the x264 git repo update into #5.  Will rebase this once that is merged.

That still leaves deciding on the version of ogg/libogg to use, fixing SRT, and the second and third bullet points from @PatTheMav 's [earlier comment](https://github.com/obsproject/obs-deps/pull/4#issuecomment-607820458.)

PatTheMav:
Alright, I guess it makes sense to continue the discussion in this project: https://github.com/obsproject/obs-studio/pull/2617#issuecomment-608139087

I successfully got FFmpeg compiled with srt-support now, including static linking against mbedtls (which needs two handcrafted files for `pkg-config`).

Do you want me to post the necessary lines/changes tomorrow so you can amend the PR? Because my changes would rely on your PR anyway.

RytoEX:
If you've gotten it to work based off of changes here, the easiest way is probably for you to PR to my branch, and then I can merge them on my branch and they'll appear here in this PR.

RytoEX:
Forgot to include a link to my branch:  https://github.com/RytoEX/obs-deps/tree/update-deps

PatTheMav:
That `libogg` fix is not enough, because you need to get that `include <inttypes.h>` back. I've created a `sed` one-liner that will take care of that.

RytoEX:
> That `libogg` fix is not enough, because you need to get that `include <inttypes.h>` back. I've created a `sed` one-liner that will take care of that.

When you say "that", do you mean staying on 1.3.3, using the commit referenced by the Windows release dependency list, or using [the latest commit from libogg's current git master](https://git.xiph.org/?p=ogg.git;a=commit;h=68ca3841567247ac1f7850801a164f58738d8df9)?

In any case, I think we need to determine which options work for our needs so that someone else can decide what the best thing to do is.  If only one option works, then that makes the decision easy.

PatTheMav:
Ah yeah, I meant that 1.3.3 also requires reintegrating the include to build `libvorbis`. `libogg` itself builds and installs fine as 1.3.3 as well as 1.3.4, but for vorbis they both miss the include in the `os_types.h`.

Monkeypatching them inside the script with this single line fixes it though:

```
sed -i '.orig' 's/#  include <sys\/types.h>/#  include <inttypes.h>/g' include/ogg/os_types.h
```

RytoEX:
Using Azure Pipelines, libvorbis seems to build fine using libogg 1.3.3.  libvorbis does not build using libogg 1.3.4.  I'm not sure why libvorbis fails to build for you with libogg 1.3.3.

Building libvorbis with libogg at their current commit also works, though it requires installing automake or using cmake+Xcode.

_Personally_, I'd prefer not to have to manually patch a dependency if it can be avoided, even if it is a one-liner.

PatTheMav:
Yeah it shouldn't be necessary on macOS 10.9+ and I have currently no idea why building on 10.15 breaks again. That could of course mean that somehow my library subsystem is broken.. 😕

RytoEX:
> Yeah it shouldn't be necessary on macOS 10.9+

Since we officially only support 10.12+, supporting builds for older than that should not be an issue.

Glad to hear you got libvorbis building with libogg.  Weird that it had issues before.

PatTheMav:
Unfortunately I spoke too soon - I don't know how or why it was able to built, but couldn't replicate it when testing the updated shell script.

But it looks like even Homebrew needs to monkey patch libogg to make it build, as this specific commit is already merged upstream but not included in official builds right now:

https://github.com/xiph/ogg/commit/c8fca6b4a02d695b1ceea39b330d4406001c03ed

RytoEX:
Yes, I've built this full set of dependencies, including libvorbis, on macOS 10.15 on Azure with both libogg 1.3.3 and [their current git master latest from their xiph.org repo](https://git.xiph.org/?p=ogg.git;a=commit;h=68ca3841567247ac1f7850801a164f58738d8df9).  Strangely, the xiph.org repo is one commit ahead of their GitHub, even if it's just a docs update.  You can see the pipelines on the branches on my fork.

I think it's safe to use the current latest git master from them, but I'd like some guidance on what to do before merging that change into this PR.

PatTheMav:
So I've just ran your deps-ogg-68ca384 branch on my machine and everything worked without error. I also compiled OBS 25.0.1 with it and have seen no issues so far.

I'd suggest updating the PR with the changes for libogg and then I can work on getting srt+mbedtls and FFmpeg to work on Mac seperately (I will try a different approach today, closely following the Homebrew build procedures).

RytoEX:
I've squashed the commits from my deps-ogg-68ca384 branch and cherry-picked the libogg commit onto this branch.

RytoEX:
@PatTheMav This has been merged.  You're good to go PR'ing your stuff for mbedTLS+SRT+FFmpeg directly to this repo.

PatTheMav:
@RytoEX Thanks, I settled on a setup yesterday, will do a PR later tonight; I still have some tests I want to pass first.

PR #13, opened 2020-04-15
Use Homebrew bundle to install missing brew dependencies

PatTheMav:
### Description
Homebrew bundle is the suggested tool to manage and bundle Homebrew dependencies. It automatically re-uses existing packages and installs required ones. Due to its design it can even unlink existing packages and side-load pinned formulas, all defined in a single Brewfile.

### Motivation and Context
Fixes issues with different Homebrew environments as it installs and updates formulas as necessary, so no fallback code in CI scripts is needed.

This also avoids CI scripts failing their workflows due to Homebrew returning with error when either a) an installed package is tried to be installed again or b) a package that needs to be unlinked isn't installed in the first place.

### How Has This Been Tested?
* Tested locally
* Tested with CI runs

### Types of changes

### Checklist:

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,0 +1,4 @@
+brew "nasm"
+brew "pkg-config"
+brew "automake"
+brew "libtool"
--- a/osx-install-tools.sh
+++ b/osx-install-tools.sh
@@ -1,1 +1,1 @@
-brew install nasm pkg-config automake
+brew bundle


Brewfile
```diff
@@ -0,0 +1,4 @@
+brew "nasm"
+brew "pkg-config"
+brew "automake"
+brew "libtool"
```
RytoEX, line 4: Leave a newline at the end of the file.

PatTheMav: Agreed, change pushed already.

PR #16, opened 2020-05-19
Update macOS minimum targets from 10.11 to 10.13

RytoEX:
### Description

Update the minimum macOS build targets from 10.11 to 10.13.

### Motivation and Context

OBS officially only supports 10.13+ because Qt 5.14.x only supports 10.13+.  Let's just make sure everything is targeting that minimum so that we (hopefully) get the benefits of targeting a newer minimum.

### How Has This Been Tested?

I haven't tested this because I don't have access to a Mac.  We should test against deps built this way, but this shouldn't really change anything.

### Types of changes

### Checklist:

### DIFF
--- a/build-package-deps-osx.sh
+++ b/build-package-deps-osx.sh
@@ -75,7 +75,7 @@ mkdir ${DEPS_DEST}/include
 mkdir ${DEPS_DEST}/lib
 
 # OSX COMPAT
-export MACOSX_DEPLOYMENT_TARGET=10.11
+export MACOSX_DEPLOYMENT_TARGET=10.13
 
 # If you need an olders SDK and Xcode won't give it to you
 # https://github.com/phracker/MacOSX-SDKs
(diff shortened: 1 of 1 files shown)


RytoEX:
The CI may fail due to git.xiph.org outages.

A thought:  could we not reuse `MACOSX_DEPLOYMENT_TARGET` in all the other places in the script?

Also, I think the main OBS repo CI similarly targets older macOS/OSX versions, and those should probably also be updated.

PatTheMav:
Good idea - had it integrated in my PR and also applied the idea to reuse it were appropriate.

RytoEX:
I still need to go make the complementary PRs for obs-studio and obs-browser so that everything is using the same minimum, and then ideally all of them should be merged around the same time.

RytoEX:
Since this is changing an obsolete file, is this even worth keeping open or merging?  Should we just remove obsolete files instead?

PatTheMav:
Yeah go ahead and do the deprecation. I actually prefer you doing that as I might be "institutionally blind" in this regard already.

RytoEX:
> Since this is changing an obsolete file, is this even worth keeping open or merging? Should we just remove obsolete files instead?

> Yeah go ahead and do the deprecation. I actually prefer you doing that as I might be "institutionally blind" in this regard already.

Submitted this removal as #52.

PR #18, opened 2020-05-24
CI: Break out dependency creation into workflow steps

PatTheMav:
### Description
- Adds version-dependent caching to all steps
- Adds Python script to create build scripts from workflows (PyYAML)
- Adds Qt as separate job in workflow with caching
- Add qtbase performance patch for QT 5.14.1 on macOS
- Adds action to push updated build scripts on workflow update
- Adds automatic release creation with created artifacts when commit is tagged

*Important note*: GitHub Actions runners have [zstd](https://facebook.github.io/zstd/) preinstalled which leads to a nasty surprise when building OBS, as the build system uses Qt's `rss` tool which is linked against `zstd` when compiled on CI. That's why it needs to be unlinked before building Qt.

Once this is merged to master, we should test the generated deps once to check if this has indeed been fixed. The caching makes this a bit tricky as the cache will contain "broken" pre-compiled builds, in that case I'll add a revision string to cache keys.

### Motivation and Context
These changes move the whole build process into Github Action workflows and require changes in this place only. When the main workflow is changed, a second workflow will pick up the event and create/update the build scripts and commits those changes back into the repository.

This also allows "pinning" of custom and/or patched Qt versions for dependencies and will only require re-compilation on version changes (cached runs take ~1.5 minutes).

### How Has This Been Tested?
* Script creation tested locally
* Created dependencies used to build/package/run OBS

### Types of changes

### Checklist:


PatTheMav:
So apparently xiph runs their stuff on Oregon State's servers and they have connectivity issues right now - might require manually restarting jobs to get all green builds. I'd advise waiting for the Qt build to be cached before doing that.. 😉

PatTheMav:
I also have a small update ready to apply https://github.com/obsproject/obs-deps/pull/16 to this PR, will force-push it later today.

RytoEX:
I confess that I don't understand the benefit of rewriting the build system this way.  All I can see from this is that we would have to change the GitHub Actions yaml files to be able to generate a build script, and this would add two dependencies (Python 3 and PyYAML) just to build the scripts so that we can build our dependencies, at least on macOS (I'm not sure if these also become required for cross-compiling Windows dependencies on Linux).

Would you please describe the expected workflow for a developer wanting to change the scripts to build new dependencies?  What benefits does this provide?

PatTheMav:
As I understand it the repository serves to build and host pre-built dependencies for common OBS builds, so building the dependencies locally is not a common use case (if it would, the lack of a public build script for the Windows dependencies would be a bigger issue, which it doesn't seem to be).

So version management and artifact generation become the primary use cases for this repository and as such those are the ones that should be made easier.

By moving the dependency composition entirely into workflow configuration it's decoupled from build scripts themselves, by separating each dependency into its own workflow step they are decoupled from other dependencies (with the exception of `ffmpeg` which requires them), which also allows caching on CI (which isn't possible with one big script), and by generating the scripts on CI upon tagging a version there is always a "stable" variant of build script+dependency configuration on the master branch.

The dependency on Python and `pyyaml` is entirely optional as a working version of the scripts will exist on the master branch once a version is tagged and those can of course be tweaked for local testing, but the actual changes need to be made to the workflow (as the main purpose of the repo is not to deliver a build script but to deliver a built dependency artifact).

(It's fine if my idea of purpose doesn't match the one of the project - it's not my decision anyway. The PR also doesn't remove the current build script, I just find it more cumbersome to manage a big script file vs. a descriptive `yml` file).

PatTheMav:
With Homebrew deprecating formula installation via URL, I've migrated Swig into the pre-built dependencies as well. This will require a minor change to the Swig Cmake module in OBS itself, but that can be adapted if this PR is merged (as the macOS build system needs to be updated there as well).

DDRBoxman:
I think putting CI first makes sense, and being able to generate scripts that can be local is a good compromise.

PR #21, opened 2020-08-05
CI: Fix Qt building imageformats plugin with Homebrew-provided deps

PatTheMav:
### Description
Hides `webp` and `libtiff` from Qt by unlinking them if installed via Homebrew, forcing Qt to use self-provided third party source code to compile both dependencies statically into its frameworks.

_Bonus Level:_ The provided build script generator was committing unusable scripts back into the repository, as an absolute checkout folder was used for script generation. The PR changes this to use the dynamically detected git checkout directory instead.

### Motivation and Context
Qt requires the `imageformats` plugin by default, which itself has dynamic libraries for tiff and webp support. If neither library is available on the system, their respective libraries are compiled into the Qt plugin library with the source code Qt provides themselves.

On CI and some local machines both these libraries might exist already (e.g. through another package that required them) and Qt will happily use those instead of the ones contained in its source repository. This leads to the unfortunate situation that the Qt libraries we ship as part of obs-deps might require these Homebrew libraries to exist on the local machine to work.

### How Has This Been Tested?
- Qt built locally with these changes, checked all plugin libraries for any other libraries linked from `/usr/local`.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/build_deps.yml
+++ b/.github/workflows/build_deps.yml
@@ -43,6 +43,7 @@ jobs:
       - name: Install Homebrew dependencies
         shell: bash
         run: |
+          brew update --preinstall
           brew bundle
       - name: Get Current Date
         shell: bash
@@ -499,6 +500,7 @@ jobs:
       - name: Install Homebrew dependencies
         shell: bash
         run: |
+          brew update --preinstall
           brew bundle
       - name: Get Current Date
         shell: bash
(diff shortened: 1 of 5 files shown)


PatTheMav:
@DDRBoxman please review and tag a new release if you have time.

Fixes https://github.com/obsproject/obs-studio/issues/3242.

PR #25, opened 2020-08-27
Build windows deps

RytoEX:
### Description

This adds scripts to cross-compile the Windows dependencies from Ubuntu.  Starting from Jim's build scripts, this gets us to a point where we can automate building the FFmpeg related dependencies on the obs-deps CI.

Obsoletes #1.

Some known issues:
* ~~The current deps package contains `zlib.dll` and `zlib.lib`.  This script generates `libzlib.dll` and `libzlib.lib`.  I couldn't figure out why this happens, so I assume Jim's scripts have a modification that I don't have here.~~
* This appears to _only_ be the FFmpeg related dependencies.  It does not build other dependencies such as cURL, FreeType, Lua/LuaJIT, Python, RNNoise, SWIG, Vulkan, or Websockets.  I'm guessing Jim builds those elsewhere.
* This doesn't follow the paradigms with GitHub workflows and script generation used by the macOS scripts.  I worked on this before that was merged, and I'd rather get this off my computer so it stops being neglected there.  If need be, it can be reworked later to fit those paradigms.

#### To-do
- [x] Fix naming of zlib DLL from libzlib.dll to zlib.dll as the current deps packages have
- [x] Moves files in `patches` to `patch` to avoid duplicate patch directories
- [ ] Address extra files produced here compared to current deps packages
- [ ] Build other deps not included here (perhaps in a follow-up PR rather than here)

### Motivation and Context

Be more transparent about our dependencies.  Automate the process of updating and deploying our dependencies.

### How Has This Been Tested?

I used the deps built by these scripts as a replacement for the current Windows deps and they seemed to work fine.  Streaming and recording worked.  Admittedly, I did this testing a few months ago, so we should probably retest more thoroughly.

To test these:
1. Download and extract the deps artifact
2. Copy the artifact's `bin` and `include` folders over the existing corresponding deps (win32 or win64)
3. Clear the OBS build folder
4. Reset the additional_install_files folder (delete the folder and then `git restore` or `git reset`)
5. Run CMake for OBS as usual to configure and generate the project
6. Open the project and build it
7. Run OBS and record, stream, or perform other actions to test

### Types of changes

### Checklist:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,3 +17,45 @@ jobs:
     inputs:
       pathtoPublish: './osx'
       artifactName: osx-deps
+
+- job: 'Build_Win32'
+  pool:
+    vmImage: 'ubuntu-18.04'
+  steps:
+  - script: ./win-install-tools.sh
+    displayName: 'Install Tools'
+  - script: ./win32.sh
+    displayName: 'Build deps'
+
+  - task: PublishBuildArtifacts@1
+    condition: ne(variables['Build.Reason'], 'PullRequest')
+    inputs:
+      pathtoPublish: './ffmpeg/ffbuild'
+      artifactName: ffmpeglogs
+
+  - task: PublishBuildArtifacts@1
+    condition: ne(variables['Build.Reason'], 'PullRequest')
+    inputs:
+      pathtoPublish: './win32'
+      artifactName: win32-deps
+
+- job: 'Build_Win64'
+  pool:
+    vmImage: 'ubuntu-18.04'
+  steps:
+  - script: ./win-install-tools.sh
+    displayName: 'Install Tools'
+  - script: ./win64.sh
+    displayName: 'Build deps'
+
+  - task: PublishBuildArtifacts@1
+    condition: ne(variables['Build.Reason'], 'PullRequest')
+    inputs:
+      pathtoPublish: './ffmpeg/ffbuild'
+      artifactName: ffmpeglogs
+
+  - task: PublishBuildArtifacts@1
+    condition: ne(variables['Build.Reason'], 'PullRequest')
+    inputs:
+      pathtoPublish: './win64'
+      artifactName: win64-deps
--- a/patch/ffmpeg/ffmpeg_flvdec.patch
+++ b/patch/ffmpeg/ffmpeg_flvdec.patch
@@ -1,0 +1,13 @@
+diff --git a/libavformat/flvdec.c b/libavformat/flvdec.c
+index b531a39adc..6b0a104f18 100644
+--- a/libavformat/flvdec.c
++++ b/libavformat/flvdec.c
+@@ -1056,7 +1056,7 @@ retry:
+             int type;
+             meta_pos = avio_tell(s->pb);
+             type = flv_read_metabody(s, next);
+-            if (type == 0 && dts == 0 || type < 0) {
++            if (type == 0 && dts == 0 || type < 0 || type == TYPE_UNKNOWN) {
+                 if (type < 0 && flv->validate_count &&
+                     flv->validate_index[0].pos     > next &&
+                     flv->validate_index[0].pos - 4 < next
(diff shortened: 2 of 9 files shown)


Fenrirthviti:
Is there any additional testing that needs to be done before this can be merged? CI looks all green, so we should be safe to merge this and start testing building obs-studio repo against these deps.

Also, per https://github.com/obsproject/obs-studio/issues/3575 can we add `--enable-libtheora` to the ffmpeg build?

Fenrirthviti:
Actually, disregard the libtheora, that should be a separate PR after this is merged to enable for both Windows and macOS so we're building the same deps for both platforms we build for.

RytoEX:
I'd love to fix the first issue regarding the naming of zlib.dll vs. libzlib.dll.  Otherwise, we _may_ have to make adjustments in the obs-studio repo for the name change.  I recall some weirdness trying to build against these deps due to the name mismatch.  My guess is that Jim's local copies of the zlib source have a custom patch for this, or there's some other step involved that I missed or messed up.

The second issue for needing to build all of the other dependencies can certainly happen later as a separate step.  I opted not to try to build every single dependency we include at first because I wanted to get these particular scripts from Jim working on CI before tackling all of the other dependencies.

The third issue is probably something that @PatTheMav would want to chime in about.  My goal when I started this was to try not to deviate too far from Jim's build scripts, but also get it running on CI.

Other than all of that, the dependencies _seemed_ to work fine for me when I last tried to use this, aside from having to adjust for the name difference for zlib/libzlib.

PatTheMav:
I'm fine with merging it as-is and then take a stab at converting this to GitHub Actions+script builder for automatic releases in a separate step. I don't have a Windows dev environment to debug this though, so I'm happy to "convert" the scripts once and see if CI likes it.

RytoEX:
For my part on the zlib thing.  I'll take another look at it this weekend.  I think I might be able to alter the configure/make files for that.

Also, it looks like I inadvertently named my patch directory "patches" instead of "patch" like @PatTheMav used for the Qt patches.  I'll fix that so that it's using "patch" instead.

RytoEX:
Force pushing a few fixes:
* Fix naming of zlib.dll
* Fix path given to dlltool for dllname option
* Make each curl retry 5 times to try to prevent failures from a single failed curl

RytoEX:
Seems to have built fine on my fork, but I haven't tested this batch yet.
https://dev.azure.com/RytoEX/obs-deps/_build/results?buildId=161&view=artifacts&type=publishedArtifacts

I still need to move the vpx patch into the correct directory.

There are still a few extra files included in my artifacts compared to the current deps:
- bin/libpng16-config
- bin/libpng-config
- bin/pngfix.exe
- bin/png-fix-itxt.exe
- bin/x264.def
- bin/x264.exe
- bin/x264.orig.def
- include/png.h
- include/pngconf.h
- include/pnglibconf.h

RytoEX:
To clarify on the zlib naming thing, I had to patch the zlib CMakeLists.txt file to add `set(CMAKE_SHARED_LIBRARY_PREFIX "")` after the project declaration.  I could probably also have done the same for `CMAKE_IMPORT_LIBRARY_PREFIX` and `CMAKE_STATIC_LIBRARY_PREFIX`.  However, I wasn't sure if there'd be any side effects, and we have `mv` in the build scripts for that anyway, so it wasn't a priority to deal with.

RytoEX:
I did a brief recording test with a few media sources and it seemed to worked fine.

To test these:
1. Download and extract the deps artifact
2. Copy the artifact's `bin` and `include` folders over the existing corresponding deps (win32 or win64)
3. Clear the OBS build folder
4. Reset the additional_install_files folder (delete the folder and then `git restore` or `git reset`)
5. Run CMake for OBS as usual to configure and generate the project
6. Open the project and build it
7. Run OBS and record, stream, or perform other actions to test

At minimum, I'd like some others to test with these before merging.  Ideally, I'd resolve the issues with extra files being generated, but I might need some more info from Jim first.  Maybe he just deletes them before making the final zips.

RytoEX:
> can we add --enable-libtheora to the ffmpeg build?
> Actually, disregard the libtheora, that should be a separate PR after this is merged to enable for both Windows and macOS so we're building the same deps for both platforms we build for.

We'd probably also have to build theora to do this, just like we do with libogg, libvpx, and libsrt.  _Probably_ not a big deal, but a bit more than just adding a compile flag.

Fenrirthviti:
Thanks, definitely reinforces that it should wait for a separate PR.

RytoEX:
Closing as this is obsoleted by #79 .

PR #26, opened 2020-08-30
CI: Add rnnoise dependency for macOS

PatTheMav:
### Description
Adds `rnnoise` pre-built library for macOS.

### Motivation and Context
Necessary for feature-parity with Windows version of OBS.

### How Has This Been Tested?
* Dependencies built locally
* OBS compiled with new dependencies
* Rnnoise available for noise canceling in filter UI

### Types of changes

### Checklist:


dodgepong:
This diff looks like it changes a lot more than adding RNNoise to the deps, is that a mistake?

PatTheMav:
@dodgepong No that's alright, those are housekeeping changes - I applied the way the git checkout for `rnnoise` is handled to `x264` as well (to use versioned checkout folders instead of just the library name). The only other change is quoting the `mbedtls` download URL.

PR #37, opened 2020-11-21
CI: Update to Qt 5.15.2 for macOS

PatTheMav:
### Description
Updates Qt to 5.15.2 as part of prebuilt dependencies.

### Motivation and Context
Prepares for milestone 26.1 as discussed in https://github.com/obsproject/obs-studio/issues/3752.

### How Has This Been Tested?
* Qt built locally
* OBS built with most recent deps and Qt
* OBS launched
* Checked VirtualCamera
* Checked Recording
* Checked Streaming (local SRT server)

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/build_deps.yml
+++ b/.github/workflows/build_deps.yml
@@ -13,8 +13,8 @@ on:
       - master
 
 env:
-  MAC_QT_VERSION: '5.14.1'
-  MAC_QT_HASH: '6f17f488f512b39c2feb57d83a5e0a13dcef32999bea2e2a8f832f54a29badb8'
+  MAC_QT_VERSION: '5.15.2'
+  MAC_QT_HASH: '3a530d1b243b5dec00bc54937455471aaa3e56849d2593edb8ded07228202240'
 
 jobs:
   macos64-main:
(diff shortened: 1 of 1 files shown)


DDRBoxman:
Leaving a note here that we need 5.15 for Apple silicon fixes 😎

RytoEX:
A bit late to be asking, I suppose, but is there a reason we load the Qt patch from a gist instead of including it in the repo?

PatTheMav:
Mainly because I wanted to use the `apply_patch` thing that expects a GitHub diff. As there was no straight diff available, this was the next simpler thing.

RytoEX:
I'm not sure I follow.  The `apply_path` utility takes a URL and a hash.  Could we not do this?

```
${{ github.workspace }}/utils/apply_patch "https://raw.githubusercontent.com/obsproject/obs-deps/master/patch/obs-issue-3799-qt.patch" "234ee6b21a008a233382c872ac097707ce83e401d3bcd5f1b244c09c0817f3aa"
```

Or we could commit the patch as its own commit and use GitHub's .patch URL?

I'm just not fond of patches living outside the repo if we can help it.  Also, not that we're building on 5.14.1 anymore, but the issue that the patch fixes affected 5.14.1 as well, but the patch is only applied to Qt versions other than 5.14.1.  We'd need a different patch to apply to 5.14.1 if we were still supporting it, but we could just rip out the 5.14.1 sections as well.

PatTheMav:
> I'm not sure I follow. The `apply_path` utility takes a URL and a hash. Could we not do this?
> 
> ```
> ${{ github.workspace }}/utils/apply_patch "https://raw.githubusercontent.com/obsproject/obs-deps/master/patch/obs-issue-3799-qt.patch" "234ee6b21a008a233382c872ac097707ce83e401d3bcd5f1b244c09c0817f3aa"
> ```
> 
> Or we could commit the patch as its own commit and use GitHub's .patch URL?
> 
> I'm just not fond of patches living outside the repo if we can help it. Also, not that we're building on 5.14.1 anymore, but the issue that the patch fixes affected 5.14.1 as well, but the patch is only applied to Qt versions other than 5.14.1. We'd need a different patch to apply to 5.14.1 if we were still supporting it, but we could just rip out the 5.14.1 sections as well.

We could and I wasn't just as smart as you just where with this idea. Love it.. :D

RytoEX:
> We could and I just wasn't as smart as you just where with this idea. Love it.. :D
> 
> We just would've needed to commit the file before committing the PR.

No worries.  We can move to that later then.  I'd much prefer keeping as much of the obs-deps data together as possible.

PR #39, opened 2020-12-15
CI: Fix incompatibilities with macOS 10.13

PatTheMav:
### Description
Fixes `lame-mp3` using a dependency symbol only available on macOS 10.14+.

### Motivation and Context
For some reason `lame-mp3` doesn't recognise the `MACOSX_DEPLOYMENT_TARGET` environment variable set by the workflow and requires it to be set explicitly when invoking `make`.

*Note:* This might mean that setting it globally doesn't work or it could just mean that lame's build process doesn't pick it up correctly. We should be on the lookout for more incompatibilities.

### How Has This Been Tested?
Dependencies were built locally and checked to not contain the offending `____chkstk_darwin` symbol.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/build_deps.yml
+++ b/.github/workflows/build_deps.yml
@@ -96,7 +96,7 @@ jobs:
         id: ffmpeg-deps-cache
         uses: actions/cache@v2.1.2
         env:
-          CACHE_NAME: 'ffmpeg-deps'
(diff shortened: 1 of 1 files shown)


PatTheMav:
@DDRBoxman Merge and tag a release pls, I need this to build artifacts on CI for our macOS 10.13 users to test if this actually fixes the crashes introduces by the prior release.

PR #49, opened 2021-01-30
CI: Build OBS dependencies for Apple Silicon Macs

nickzman:
### Description

This change updates the build scripts so that, when they are run on an Apple Silicon Mac, they will build the dependencies for an Apple Silicon Mac.

All of the dependencies build without any changes, except for libvpx and Qt. libvpx 1.9.0 does not support building for arm64 on macOS, but the master branch code does, so for the time being, we have to clone and then build from the master branch. Qt needs to be coerced to build for arm64 using the variable QMAKE_APPLE_DEVICE_ARCHS.

### Motivation and Context

Apple is switching CPU architectures. In a few years, they won't be shipping or supporting Intel Macs anymore, so it's time to start producing a build for the new Apple Silicon Macs.

### How Has This Been Tested?

I tested my changes on my Apple Development Platform running Xcode 12.4. Both scripts complete successfully, and running `lipo -info` on the resulting binaries indicate that they are built for arm64.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/build_deps.yml
+++ b/.github/workflows/build_deps.yml
@@ -185,19 +185,32 @@ jobs:
         shell: bash
         working-directory: ${{ github.workspace }}/CI_BUILD
         run: |
-          ${{ github.workspace }}/utils/safe_fetch "https://github.com/webmproject/libvpx/archive/v${{ env.LIBVPX_VERSION }}.tar.gz" "${{ env.LIBVPX_HASH }}"
-          mkdir -p ./libvpx-v${{ env.LIBVPX_VERSION }}
-          tar -xf v${{ env.LIBVPX_VERSION }}.tar.gz
+          # Note: libvpx 1.9.0, which is the most current release when this was written,
+          # doesn't support building for arm64 on macOS, but the latest development code
+          # supports building for arm64 on macOS. So, on Apple Silicon Macs, we have to clone
+          # the latest code instead of downloading an archive. This can be revised when the
+          # next version of libvpx comes out.
+          if [ `arch` = "arm64" ]; then
+            git clone "https://chromium.googlesource.com/webm/libvpx" libvpx-${{ env.LIBVPX_VERSION }}
+          else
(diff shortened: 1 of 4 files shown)


DDRBoxman:
These changes actually need to be in 
https://github.com/obsproject/obs-deps/blob/master/.github/workflows/build_deps.yml

We generate the build scripts from that:
https://github.com/obsproject/obs-deps/blob/master/.github/workflows/create_scripts.yml

nickzman:
Thanks; I didn't know about that file until now.

nickzman:
Any news on the review? I'd like to do whatever I can to bring OBS to native ARM64 on macOS, and getting the dependencies built is step one.

DDRBoxman:
looks good, would you mind setting up the GitHub actions build matrix to trigger this for x86 and arm?

looks like you can use the token strategy for each 
https://stackoverflow.com/a/59310425

DDRBoxman:
You'll probably want to stick the "arm" or "x86" tag in filenames of these releases:
https://github.com/obsproject/obs-deps/blob/master/.github/workflows/build_deps.yml#L707

nickzman:
Hmm, I think that's going to be a problem. Looking through [the documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on), it appears that Github has no macOS/ARM64 runners, nor does it allow self-hosted macOS/ARM64 runners, and the scripts need to be run on an ARM64 Mac in order to produce ARM64 dependencies.

So either the scripts need to be rewritten to cross-compile for ARM64 & we hope that they are using the latest Xcode release, or the dependencies need to be built and uploaded manually. For the time being, I am kind of hoping they could be uploaded manually...

IsaacMarovitz:
Hmmm, whenever I try to build M1 with this patch, ffmpeg fails to build.

```
  + ffmpeg-4.2.3.tar.xz downloaded successfully and passed hash check
libvpx enabled but no supported decoders found

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "ffbuild/config.log" produced by configure as this will help
solve the problem.
  + ERROR during build step: Build dependency ffmpeg
```

PR #51, opened 2021-03-22
CI: Use 'uname -m' for QMAKE_APPLE_DEVICE_ARCHS

SCG82:
### Description

Use `uname -m` instead of `arch` to set `QMAKE_APPLE_DEVICE_ARCHS` when building Qt.

### Motivation and Context

Currently, `QMAKE_APPLE_DEVICE_ARCHS` is set using `arch`, which returns `i386` for Intel-based machines. `QMAKE_APPLE_DEVICE_ARCHS` expects `x86_64`, which `uname -m` returns. Both `arch` and `uname -m` return `arm64` on Apple M1 machines.

### How Has This Been Tested?

With `QMAKE_APPLE_DEVICE_ARCHS` being set by `arch` (result `i386`), Qt 5.15.2 failed to build on my 2018 Mac Mini (6-core) running macOS 11.2. Setting `QMAKE_APPLE_DEVICE_ARCHS` using `uname -m` (result `x86_64`) resolved this issue. 

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/build_deps.yml
+++ b/.github/workflows/build_deps.yml
@@ -685,7 +685,7 @@ jobs:
             -skip qtscript -skip qtscxml -skip qtsensors -skip qtserialbus -skip qtspeech \
             -skip qttranslations -skip qtwayland -skip qtwebchannel -skip qtwebengine -skip qtwebglplugin \
             -skip qtwebsockets -skip qtwebview -skip qtwinextras -skip qtx11extras -skip qtxmlpatterns \
-            QMAKE_APPLE_DEVICE_ARCHS=`arch`
+            QMAKE_APPLE_DEVICE_ARCHS=$(uname -m)
           make -j${{ env.PARALLELISM }}
           make install
 


build-script-macos-02.sh
```diff
@@ -200,7 +200,7 @@ build_06_build_dependency_qt() {
       -skip qtscript -skip qtscxml -skip qtsensors -skip qtserialbus -skip qtspeech \
       -skip qttranslations -skip qtwayland -skip qtwebchannel -skip qtwebengine -skip qtwebglplugin \
       -skip qtwebsockets -skip qtwebview -skip qtwinextras -skip qtx11extras -skip qtxmlpatterns \
-      QMAKE_APPLE_DEVICE_ARCHS=`arch`
+      QMAKE_APPLE_DEVICE_ARCHS="$(uname -m)"
```
RytoEX, line 203: Don't modify the shell scripts.  They are automatically rebuilt using the `create_scripts.yml` workflow.

RytoEX (changes requested):

build-script-macos-02.sh
SCG82: Modified `build_deps.yml` (and found an extra `=` in the `CXX` assignment)

.github/workflows/build_deps.yml
```diff
@@ -229,7 +229,7 @@ jobs:
             brew install gcc || true
             CC="/usr/local/bin/gcc"
             LD="/usr/local/bin/gcc"
-            CXX=="/usr/local/bin/g++"
+            CXX="/usr/local/bin/g++"
```
RytoEX, line 232: This CXX change should be a separate commit and/or a separate PR.  Otherwise, blame will point to this commit, and reverting this commit will revert both changes.

RytoEX (changes requested):

.github/workflows/build_deps.yml
SCG82: done

PR #60, opened 2021-06-14
CI: Add per-dependency build scripts and update CI workflow

PatTheMav:
### Description

This PR adds universal binary support for all pre-compiled OBS dependencies. This is achieved via the following changes:

* Each dependency's build step has been moved into its separate build script
* Build scripts use a pre-defined template for patching/building/preparing to keep necessary customisations per-dependency low
* Each dependency can be built standalone (via its own script) and allows explicit specification of the build architecture (`x86_64`, `arm64`, or `universal`), the host's architecture is used by default
* Dylib dependency have their absolute paths replaced by `@rpath` references, pre-existing `LC_RPATH` values are removed before packaging or creation of universal binaries via `lipo`.
* Separate scripts to generate universal binaries given `arm64` and `x86_64` build trees
* Patches for dependencies are now kept as patch files within the repository

There are still warts with this PR (mainly the macOS version baked into some libraries is not correct) and the workflow itself needs to be tested on CI of course.

### Motivation and Context
This change allows OBS to be built and packaged as either a universal or platform-specific build and also limits the housekeeping for each dependency to its own build script, reducing the maintenance and update effort.

### How Has This Been Tested?
* Builds for all three variants ((`x86_64`, `arm64`, and `universal`) built locally with fresh as well as existing temporary build directory 
* OBS built with x86 libraries
* OBS built with universal libraries (detected as Apple M1 binary on an M1 MacBook Air)

### Types of changes

### Checklist:


PatTheMav:
As always CI breakage with issues that never occurred on local systems, will have to investigate why Qt ignores the c++17 switch and x264 all of a sudden doesn't want to cross-compile arm64 anymore.

@Developer-Ecosystem-Engineering feel free to take a look.

utils/build_script_generator.py
```diff
@@ -2,168 +2,83 @@
...
+                    if key
+                    not in [
+                        "MACOSX_DEPLOYMENT_TARGET",
+                        "FFMPEG_REVISION",
+                        "BLOCKED_FORMULAS",
+                    ]
+                }
+
+                # Use preserved item order in dicts with Py 3.7
+                dependencies = dict.fromkeys(
```
PatTheMav, line 40: @derrod lmk if you know of an easier pattern to get get pairs of `A_VERSION` and `A_HASH` into a single list containing only `A` while also preserving order.. 😬

derrod: Wouldn't a good old for-loop be simpler?

Something like
```python
dependency_strings = []

for key in filtered_data.keys():
    prefix = key.partition('_')[0]
    dependency_strings.append(
        f'    "{prefix.lower()}'
        f' {environment_data.get(f"{prefix}_VERSION", "")}'
        f' {environment_data.get(f"{prefix}_HASH", "")}"'
    )
```

A normal python list preserves order anyway so I'm not entirely sure why you're essentially creating a list from the `dict_keys` view, only to then turn that into a dict again, which then gets used in a `dict_keys` view again. The list from `[key.split("_")[0] for key in filtered_data.keys()]` could be used directly.

derrod: Or wait, the issue is that things aren't unique, right?

PatTheMav: The double conversion eliminates duplicates, that's the trick. By creating a dict from the keys minus the suffix, I eliminate the duplicates, but still preserve order. `set(dict.keys()` shuffles the order, so that was out of the question.. 😂

derrod: In that case the only thing I really see is just using a generator instead of creating a temporary list, and combining the two into one big mess. A for-loop over a dictionary will also iterate over keys by default so you don't need to explicitly do `.keys()`

```python
dependency_strings = [
    (
        f'    "{key.lower()}'
        f" {environment_data.get(f'{key}_VERSION', '')}"
        f" {environment_data.get(f'{key}_HASH', '')}\""
    )
    for key in dict.fromkeys(key.partition('_')[0] for key in filtered_data)
]
```

PatTheMav:
@RytoEX @DDRBoxman - if we could get this reviewed, merged and released first, I could adapt pending PRs to these new deps and add preparations for universal/multi-arch builds. I got a branch prepared for that, but that's obviously based on my rework PR for now.

RytoEX:
> 
> 
> @RytoEX @DDRBoxman - if we could get this reviewed, merged and released first, I could adapt pending PRs to these new deps and add preparations for universal/multi-arch builds. I got a branch prepared for that, but that's obviously based on my rework PR for now.

I'd looked over the PR quickly when it was submitted and it seemed fine at a glance, though I haven't gone over it in depth.  I'm much in favor of this approach over the YAML -> script approach.  I think my one request at the moment would be that if patches are OS-agnostic, that they be located in a global patches directory instead of sequestered in OS-specific sub-directories.

PatTheMav:
> I think my one request at the moment would be that if patches are OS-agnostic, that they be located in a global patches directory instead of sequestered in OS-specific sub-directories.

I agree, but as we have no other OSes or even "global" patches right now we should keep this in mind when we add Windows deps building.

.github/workflows/create_scripts.yml
```diff
@@ -3,10 +3,9 @@ name: build_macos
 on:
   push:
     paths:
-      - '.github/workflows/build_deps.yml'
+      - '.github/workflows/master.yml'
```
RytoEX, line 6: There is no workflow with this name.  Is that intentional?  What does the `create_scripts.yml` do now, anyway?

RytoEX:
> 
> 
> > I think my one request at the moment would be that if patches are OS-agnostic, that they be located in a global patches directory instead of sequestered in OS-specific sub-directories.
> 
> I agree, but as we have no other OSes or even "global" patches right now we should keep this in mind when we add Windows deps building.

The QTBUG-74606.patch affects Windows.  We build with it currently.

.github/workflows/create_scripts.yml
PatTheMav: Right, it should say `main` - will push a fix in a bit.

The script now reads the dependency config in the workflow and updates the global build script, so we don't have to update the versions/hashes in two files.

The global build script is convenience for local builds, so I thought this would make maintenance convenient for us as well as keeping that script up2date whenever we do a release.

RytoEX: Ah, so now it reads https://github.com/obsproject/obs-deps/pull/60/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R23-R56 and updates https://github.com/obsproject/obs-deps/pull/60/files#diff-c1f2aefe3a1708caf92d65020c7b456155cb3c8b8f7dbc69a8f0cb1076130893R25-R42 ?  Is that about it?  Does that necessitate having the exact same deps for each OS we build for?

PatTheMav: Not in that specific case, because the script uses the values set as `env` in the OS-specific build job. So if you specify different versions for Windows, it wouldn't change a thing.

It might trip up the standalone scripts' fallbacks (as they are read from the workflow and take the first value they encounter), but I'll pass that bridge once we get there. I'll probably have to use more complex parsing in Powershell anyway so that becomes a non-issue.

RytoEX: Sounds good for now.

PatTheMav:
> > > I think my one request at the moment would be that if patches are OS-agnostic, that they be located in a global patches directory instead of sequestered in OS-specific sub-directories.
> > 
> > 
> > I agree, but as we have no other OSes or even "global" patches right now we should keep this in mind when we add Windows deps building.
> 
> The QTBUG-74606.patch affects Windows. We build with it currently.

Got it, will relocate it together with the fix for the script creation workflow then.

.github/workflows/create_scripts.yml
PatTheMav: Should be fixed with a recent force-push.

README.md
```diff
@@ -2,33 +2,39 @@
...
+## macOS (10.13+ for x86_64, 11.0 for arm64)
 
 | lib | git commit | version |
 | :--- | :---: | :---: |
 |libpng|[Sourceforge](https://downloads.sourceforge.net/project/libpng/libpng16/1.6.37/libpng-1.6.37.tar.xz)|1.6.37|
-|libopus|[Oregon State University FTP](https://ftp.osuosl.org/pub/xiph/releases/opus/opus-1.3.1.tar.gz)|1.3.1|
-|libogg|[xiph.org](https://downloads.xiph.org/releases/ogg/libogg-1.3.4.tar.gz)|1.3.4|
-|librnnoise|[90ec41e](https://github.com/xiph/rnnoise/commit/90ec41ef659fd82cfec2103e9bb7fc235e9ea66c)||
+|libopus|[Oregon State University FTP](https://github.com/xiph/opus/tree/dfd6c88aaa54a03a61434c413e30c217eb98f1d5)|Master branch|
+|libogg|[xiph.org](https://github.com/xiph/ogg/releases/download/v1.3.5/libogg-1.3.5.tar.xz)|1.3.5|
```
RytoEX, line 11: This link is not for xiph.org, but rather their GitHub mirror.

.github/workflows/main.yml
```diff
@@ -0,0 +1,587 @@
...
+
+          find . \( -type f -or -type l \) \( -name "*.la" -or -name "*.a" \) | xargs rm
+          find ./lib -mindepth 1 -maxdepth 1 -type d | xargs rm -rf
+          rm -rf ./bin
+          rm -rf ./share
+          cp -R ${{ github.workspace }}/licenses .
+
+          XZ_OPT=-T0 tar -cJf "${FILE_NAME}" *
+
+          mv ${FILE_NAME} ..
```
RytoEX, line 326: Would this make more sense as its own script instead of raw script in yaml?

RytoEX (changes requested):
Just a quick visual review.  Nothing really stood out other than some stuff in the README.  Someone else will have to handle actually doing test runs on the build scripts.  While it'd be good to test these locally, ultimately we'll also have see how it works after merging and have someone test the resulting dependencies.  Unfortunately, I can do neither of these steps.

I'm not sure if I like having everything in a "CI" directory, but I can't think of an alternative that is better, so I'm not going to fuss about it.  A "scripts" directory doesn't sound much better, and if it's going to contain patches, it's also a bit confusing.  Conceptually, I'd prefer if the patches were in a root-level subdirectory, but maybe in practice that doesn't work as well.  I'll leave that to you.

.github/workflows/main.yml
PatTheMav: It is implemented in the packaging scripts but those contain much more convenience functionality that wasn't needed here.

.github/workflows/main.yml
```diff
@@ -0,0 +1,587 @@
...
+    runs-on: [macos-11]
+    strategy:
+      matrix:
+        arch: ['x86_64', 'arm64']
+    env:
+      MACOSX_DEPLOYMENT_TARGET: '10.13'
+      CACHE_REVISION: '02'
+      LIBPNG_VERSION: '1.6.37'
+      LIBPNG_HASH: '505e70834d35383537b6491e7ae8641f1a4bed1876dbfe361201fc80868d88ca'
+      LIBOPUS_VERSION: '2021-05-12'
```
RytoEX, line 27: I understand why it's appealing to set the version to 2021-05-12.  However, if you run the `update_version` script included in the xiph/opus repo with this commit checked out, it will output a `package_version` file with "1.3.1-93-gdfd6c88a" as its set package version.  It makes more sense to use the authoritative version info rather than making up our own.  We should either leave this at 1.3.1, or use the version info generated by their script.

.github/workflows/main.yml
```diff
@@ -0,0 +1,587 @@
...
+      SPEEXDSP_HASH: 'd7032f607e8913c019b190c2bccc36ea73fc36718ee38b5cdfc4e4c0a04ce9a4'
+      LIBJANSSON_VERSION: '2.13.1'
+      LIBJANSSON_HASH: 'f4f377da17b10201a60c1108613e78ee15df6b12016b116b6de42209f47a474f'
+      LIBLUAJIT_VERSION: '2.1'
+      LIBLUAJIT_HASH: 'ec6edc5c39c25e4eb3fca51b753f9995e97215da'
+      LIBFREETYPE_VERSION: '2.10.4'
+      LIBFREETYPE_HASH: '86a854d8905b19698bbc8f23b860bc104246ce4854dcea8e3b0fb21284f75784'
+      LIBRNNOISE_VERSION: '2020-07-28'
+      LIBRNNOISE_HASH: '90ec41ef659fd82cfec2103e9bb7fc235e9ea66c'
+      BLOCKED_FORMULAS: 'ant gradle maven selenium-server-standalone llvm gcc postgresql openjdk sox libsndfile flac libvorbis opusfile libogg php gd freetype fontconfig webp libpng lame libtiff opus'
```
RytoEX, line 57: Should this be sorted alphabetically so there's some order to it?

README.md
```diff
@@ -2,33 +2,39 @@
...
-## macOS (10.13+)
+## macOS (10.13+ for x86_64, 11.0 for arm64)
 
 | lib | git commit | version |
 | :--- | :---: | :---: |
 |libpng|[Sourceforge](https://downloads.sourceforge.net/project/libpng/libpng16/1.6.37/libpng-1.6.37.tar.xz)|1.6.37|
-|libopus|[Oregon State University FTP](https://ftp.osuosl.org/pub/xiph/releases/opus/opus-1.3.1.tar.gz)|1.3.1|
-|libogg|[xiph.org](https://downloads.xiph.org/releases/ogg/libogg-1.3.4.tar.gz)|1.3.4|
-|librnnoise|[90ec41e](https://github.com/xiph/rnnoise/commit/90ec41ef659fd82cfec2103e9bb7fc235e9ea66c)||
+|libopus|[Github](https://github.com/xiph/opus/tree/dfd6c88aaa54a03a61434c413e30c217eb98f1d5)|Master branch|
```
RytoEX, line 10: See comment above about specific opus version.

README.md
```diff
@@ -39,23 +45,29 @@ git clone https://github.com/obsproject/obs-deps.git
...
 
 ## Contributing
 
-* Do not make changes to the `build-script-{macos/ubuntu/windows}-{num}.sh` scripts, which are generated by `.github/workflows/create_scripts.yml`
-* Instead, edit `.github/workflows/build_deps.yml`
\ No newline at end of file
+* Add/edit seperate build scripts for every dependency in the `CI/[OPERATING SYSTEM/` directory
+* For new dependencies:
+    * Create the `sha256sum` of the downloaded dependency archive
+    * Add the dependency version as `[DEPENDENCY_NAME]_VERSION` and the downloaded archive hash as `[DEPENDENCY_NAME]_HASH` to the Github actions workflow in `.github/workflows/main.yml` as well as the main build script in `CI/build-deps-macos.sh`.
```
RytoEX, line 68: Nitpick:  GitHub

This improper capitalization exists elsewhere in this PR as well (dependency version table), but in places that already exist.  Those other instances can easily be corrected separately.

README.md
```diff
@@ -39,23 +45,29 @@ git clone https://github.com/obsproject/obs-deps.git
...
+* For existing dependencies:
+    * Always update the `sha256sum` of the updated dependency archive as well as the version
+* Patches need to be placed either in the patches directory (if applied for all OS') or inside the patches directory for a specific OS
```
RytoEX, line 71: `OS'` doesn't need an apostrophe by itself.  If it were `OS's`, then maybe, but not required, according to APA guidelines.  I'll leave the choice between `OS` vs. `OSs` vs. `OS's` up to you.

RytoEX (changes requested):
Mostly nitpicks.

I'd like to see some additional confirmation that the deps produced from this PR work fine for building OBS on macOS.  This PR is generating artifacts, so anyone with a Mac who can build OBS should be able to help with this.  Keeping track of what version of macOS and what host arch was used would be helpful.

Developer-Ecosystem-Engineering:
> As always CI breakage with issues that never occurred on local systems, will have to investigate why Qt ignores the c++17 switch and x264 all of a sudden doesn't want to cross-compile arm64 anymore.
> 
> @Developer-Ecosystem-Engineering feel free to take a look.

Was able to build + run against obs-studio/master using the arm64 dependencies generated with this change 

- MacBook Air M1
- macOS 11.2.3 
- Clang 12.0.5

Was able to capture from built in camera, record using the default .mkv and .mp4. 

obs-apple-silicon

.github/workflows/main.yml
PatTheMav: Sorted by build order which is also in parts dependant on quirks in how FFmpeg dependencies need to be built (i.e. MbedTLS and SRT).

OsvaldoRino (approved):

CI/build-qt-macos.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+##############################################################################
+# macOS dependencies build script
+##############################################################################
+#
+# This script pre-compiles all dependencies required to build OBS
```
DDRBoxman, line 7: Incorrect description

DDRBoxman (changes requested):

PR #63, opened 2021-09-17
Add CEF patch file

Fenrirthviti:
### Description
This adds the CEF patch file used for OSR support on CEF version 4430 that OBS
requires for hardware acceleration to work properly. It is being committed here
for tracking and transparency, and if we need any future patches they can be
maintained here instead of pulled from a gist from a link in a PR that is
closed.

### Motivation and Context
We apply this patch to the builds of CEF that we ship with OBS. Maintaining a copy of the patch here allows for a more transparent build process, and negates having to rely on the patch from a comment on a PR that is not likely to ever be merged.

### How Has This Been Tested?
Patch is used with shipped CEF builds for OBS Studio.

### Types of changes

### Checklist:

- [X] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [X] I have included updates to all appropriate documentation.

### DIFF
--- a/patches/cef/viz_osr_d3d_4430.patch
+++ b/patches/cef/viz_osr_d3d_4430.patch
@@ -1,0 +1,1676 @@
+diff --git a/BUILD.gn b/BUILD.gn
+index ec53103d2..31c77e9b7 100644
+--- a/BUILD.gn
++++ b/BUILD.gn
+@@ -1880,6 +1880,7 @@ if (is_mac) {
+     frameworks = [
+       "AppKit.framework",
+       "OpenGL.framework",
++      "IOSurface.framework",
+     ]
+     defines = [
+       "CEF_USE_SANDBOX",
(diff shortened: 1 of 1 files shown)


RytoEX:
This adds a new root-level directory `patches`, instead of putting a file into the existing `CI/patches` directory.

While I pointed out that I wasn't fond of using `CI/patches`, I'd like it if we could be consistent one way or another.  That means we either add patches to the existing `CI/patches`, or we decide to switch to using `patches`.

Fenrirthviti:
Ah, my mistake I wasn't aware there was already a patches folder. I'll move it there. It also sounds like there might be a patch for this patch that will need to be included as well.

This was more or less a rage PR from the confusing ecosystem that is our CEF builds. :)

Fenrirthviti:
https://github.com/obsproject/cef

closing this

PR #65, opened 2021-10-11
Block SBT on macOS

RytoEX:
### Description

Block SBT on macOS.

SBT was recently added to the macOS 11 runners.

### Motivation and Context

Fix broken macOS CI.

### Types of changes

### Checklist:


PatTheMav:
We "ruin" the Homebrew environment of the runner one way or another, so it should be fine if we break dependencies for packages we don't use anyway.

PR #69, opened 2021-10-15
Fix target OS versions for FFmpeg libraries

PatTheMav:
### Description
Fixes the target macOS versions for FFmpeg libraries.

### Motivation and Context
While many projects use the "internal" macOS version number as part of their target definitions, FFmpeg uses the release version numbers, so `10.13` has to be used instead of `17`, otherwise the linker will complain about FFmpeg libraries targeting a newer macOS version than the binary that is being linked to.

### How Has This Been Tested?
* Test locally

### Types of changes

### Checklist:


CI/macos/build_ffmpeg.sh
```diff
@@ -72,8 +72,8 @@ _build_product() {
             --enable-cross-compile --enable-shared --disable-static --enable-pthreads \
             --shlibdir="${BUILD_DIR}/lib" --pkg-config-flags="--static" --prefix="${BUILD_DIR}" --enable-rpath \
             --host-cflags="-I${BUILD_DIR}/include" --host-ldflags="-L${BUILD_DIR}/lib"  \
-            --extra-ldflags="-target x86_64-apple-macos${DARWIN_TARGET} -L${BUILD_DIR}/lib -lstdc++" \
-            --extra-cflags="-fno-stack-check -target x86_64-apple-macos${DARWIN_TARGET} -I${BUILD_DIR}/include" \
+            --extra-ldflags="-target x86_64-apple-macos10.13 -L${BUILD_DIR}/lib -lstdc++" \
+            --extra-cflags="-fno-stack-check -target x86_64-apple-macos10.13 -I${BUILD_DIR}/include" \
```
RytoEX, line 76: Could we use `MACOSX_DEPLOYMENT_TARGET` or some equivalent here?  Not sure about the arm64 branch.

PR #71, opened 2021-10-16
Add function to check git config

RytoEX:
### Description

Git requires a user to be set up before certain operations will work (e.g., cherry-pick).

The `git cherry-pick` operation requires that a git user (user.email and user.name) be set up.  If one is not set up (like on CI), let's set one up to use.

`git config --get` will produce an error if the key cannot be found or it is not set, so I have temporarily `set +e` around these calls.  If there is a preferred alternative, such as appending ` || true` to the calls, let me know.

### Motivation and Context

I want to be able to use `git cherry-pick`.

### How Has This Been Tested?

Tested locally on Ubuntu.

### Types of changes

### Checklist:

### DIFF
--- a/CI/include/build_support.sh
+++ b/CI/include/build_support.sh
@@ -77,6 +77,32 @@ check_ccache() {
     fi
 }
 
+check_git() {
+    step "Check git..."
+    if git --version >/dev/null 2>&1; then
+        info "Git available"
+
+        info "Check git config for user..."
+        git_user_email=$(git config --get user.email)
+        if [ -z "$git_user_email" ]; then
+            info "Set git user.email..."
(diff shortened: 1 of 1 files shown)


RytoEX:
After further testing, while this works if called from within a build script such as `CI/build-deps-macos.sh`, it still causes an error if called from a component build script.  Setting it to draft until that's worked out.

tytan652:
On my AUR packages when I need to cherry-pick I add `--no-commit`, but on this I don't know if it's a good idea for this project.

RytoEX:
Going to merge this.

A good future improvement would be to make this better handle any errors it emits without having to wrap calls in toggles of the error mode.
```
set +eE
check_git
set -eE
```

PR #78, opened 2021-11-09
Add support for git sparse-checkout

RytoEX:
### Description

This commit adds support for a component build script to pass the "set \<sparse-path>" string to github_fetch via _build_setup_git.

### Motivation and Context

Sparse checkouts can drastically reduce the space necessary to checkout a repository if we only need to clone specific parts of the repository.

Includes #71 because the changes were too close together to commit wholly separately.

### How Has This Been Tested?

Have used locally and on GHA while testing the Windows deps branch.

### Types of changes

### Checklist:


RytoEX:
Merging this.

Good future improvements would be:
* turn `is_gte` into a `version_compare` function
* use return codes instead of echoing strings

PR #73, opened 2021-10-20
Prevent EOL conversions for licenses

RytoEX:
### Description

Prevent EOL conversions for licenses

License files should probably be committed as-is without trying to force the EOL marker to adhere to any particular format.  This change to `.gitattributes` should exclude all files in all sub-folders of `/licenses` from all end-of-line conversion.[1]

[1]: https://git-scm.com/docs/gitattributes#_text

### Motivation and Context

Prevent unnecessary changes to license files.

### Checklist:

- [ ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [ ] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #79, opened 2021-11-10
Add Windows dependencies build scripts and workflows

RytoEX:
### Description

Build all Windows dependencies present in the current dependencies2019.zip file.  This PR is intended to produce a direct replacement for that file.

This PR includes commits from the following PRs, which I hope can be addressed before this if there are no issues:
* #71 
* #73 
* #78 

This PR obsoletes PR #25.

Caveats:
* Currently, the native Windows build scripts are set to try to install dependencies via Chocolatey (and Chocolatey itself) by default.  This will attempt to install 7-Zip, CMake, and pyenv-win and add CMake to your PATH.  If you have these installed beforehand, and CMake is in your PATH, it will use your existing installations.  I plan to swap this behavior in a later iteration to have the default be to check for these build dependencies, but require the user to actively specify that they want the tools installed via Chocolatey, but I do not want to hold this up on that change.
* If the latest version of CMake is already installed on your PC but is not in your PATH, the installer will fail to make this change due to the nature of MSI installers (which CMake uses).  To fix this, you can either add CMake to your PATH manually, or uninstall and reinstall CMake manually.
* Some Windows native build scripts currently require a VS2019 installation to be present on the host system.  VS2015, VS2017, and VS2022 are not currently supported.  VS2022 support will be added later.
* Ubuntu 20.04's base apt repos have broken libc6-dev packages, so it is not recommended to try to cross-compile using that as a host system at this time.  Ubuntu 16.04, 18.04, 21.04, and 21.10 all seem to be fine.  See https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1926918.  Ubuntu 20.04 on GitHub Actions seems to have avoided this issue.

### Motivation and Context

Get all of our dependencies building with scripts that can run in most environments.  Document what patches and build flags are used.

### How Has This Been Tested?

On Windows 10 21H1 (19043.1288), I have tested builds produced from these scripts by doing the following:
* recording with NVENC, QSV, and x264
* streaming to Twitch with NVENC, QSV, and x264
* streaming to YouTube RTMPS with NVENC
* streaming to YouTube HLS with NVENC
* Display Capture (WGC only)
* Window Capture (WGC and BitBlt)
* Game Capture
* Color Source
* Image Source (PNG, JPG, animated GIF)
* Video Capture Device
* browser source loading the default page
* webm stingers
* Media Source with local MP4 and FLV files
* Media Source with SRT feeds from Larix Broadcaster and Larix Screencaster
* the Lua Clock script and source
* VLC Source with multiple files, including source toolbar / media controls
* Text Sources (both GDI+ and FreeType2)
* Audio Filters (most of them, including Speex and RNNoise filters)
* Chroma Key Filter on a Media Source
* Undo/Redo

### Types of changes

### Checklist:


CI/windows/build_ffmpeg.sh
```diff
@@ -0,0 +1,117 @@
...
+        c112fae6603f8be33cf1ee2ae390ec939812f473 \
+        86a7b77b60488758e0c080882899c32c4a5ee017 \
+        7cc7680a802c1eee9e334a0653f2347e9c0922a4 \
+        449e984192d94ac40713e9217871c884657dc79d \
+        290a35aefed250a797449c34d2f9e5af0c4e006a \
+        6e95ce8cc9ae30e0e617e96e8d7e46a696b8965e \
+        e9b35a249d224b2a93ffe45a1ffb7448972b83f3 \
+        7c59e1b0f285cd7c7b35fcd71f49c5fd52cf9315 \
+        86f5fd471d35423e3bd5c9d2bd0076b14124faee \
+        fb0304fcc9f79a4c9cbdf347f20f484529f169ba
```
notr1ch, line 47: It would be nice if these were commented with the commit title so its easy to see what they are.

RytoEX: I'll see if I can do that quickly.

CI/include/build_support_windows_cross.sh
```diff
@@ -0,0 +1,192 @@
...
+    COLOR_ORANGE=""
+    COLOR_RESET=""
+fi
+
+## DEFINE UTILITIES ##
+install_tools() {
+    sudo apt ${QUIET:+--quiet} -y install automake cmake curl git libtool mingw-w64 mingw-w64-tools pkg-config wget
+
+    curl -L -O https://www.nasm.us/pub/nasm/releasebuilds/2.15.01/nasm-2.15.01.tar.xz
+    tar -xf nasm-2.15.01.tar.xz
```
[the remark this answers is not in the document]
RytoEX: It's hash checked in the build script itself because we're just using prebuilt versions of NASM and we need both 32-bit and 64-bit, for which the hashes differ.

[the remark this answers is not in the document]
RytoEX: Hash should now be checked for this download.

CI/windows/build_ffmpeg.sh
RytoEX: A comment block with the output of `git log --oneline` that matches this commit order has been added.

.github/workflows/windows_deps.yml
```diff
@@ -0,0 +1,649 @@
...
+        shell: pwsh
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
+
+      - name: 'Check for Github Labels'
+        if: github.event_name == 'pull_request'
+        shell: bash
+        run: |
+          if test -n "$(curl -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -s "${{ github.event.pull_request.url }}" | jq -e '.labels[] | select(.name == "Seeking Testers")')"; then
```
PatTheMav, line 279: You could use Powershell native code here:
```
          $LabelFound = try { (Invoke-RestMethod -Authentication 'Bearer' -Token (ConvertTo-SecureString '${{ secrets.GITHUB_TOKEN }}' -AsPlainText) -Uri "${{ github.event.pull_request.url }}" -UseBasicParsing).labels.name.contains("Seeking Testers") } catch { $false }
          Write-Output "SEEKING_TESTERS=$(if( $LabelFound -eq $true ) { 1 } else { 0 })" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf8 -Append
```

CI/package-deps-windows.ps1
```diff
@@ -0,0 +1,159 @@
...
+        # Copy cross-compiled deps first
+        Write-Step "Copy cross-compiled $($Package.WinArch)/$($Package.Arch) files..."
+        Copy-Item -Path "${CrossDir}\bin" -Destination "${FinalDir}" -Recurse
+        Copy-Item -Path "${CrossDir}\include" -Destination "${FinalDir}" -Recurse
+        Copy-Item -Path "${CrossDir}\licenses" -Destination "${FinalDir}" -Recurse
+
+        Write-Step "Clean up cross-compiled $($Package.WinArch)/$($Package.Arch) files..."
+        # Remove unneeded files before copying native-compiled deps
+        # Make sure symlinks still exist before trying to remove them
+        Remove-Item -Path "${FinalDir}\bin\libmbedtls.dll" -Force
```
PatTheMav, line 94: I wonder if it might be simpler to add the file paths to an array and just pipe it to `Remove-Item`:

```
@(
    "${FinalDir}\bin\libmbedtls.dll",
    "${FinalDir}\bin\libmbedx509.dll",
    [...]
) | Remove-ItemIfExists -Force
```

CI/package-deps-windows.ps1
```diff
@@ -0,0 +1,159 @@
...
+        Remove-Item -Path "${FinalDir}\bin\mbedx509.lib" -Force
+        Remove-Item -Path "${FinalDir}\bin\pngfix.exe" -Force
+        Remove-Item -Path "${FinalDir}\bin\png-fix-itxt.exe" -Force
+        Remove-ItemIfExists "${FinalDir}\bin\srt-ffplay"
+        Remove-Item -Path "${FinalDir}\bin\x264.def" -Force
+        Remove-Item -Path "${FinalDir}\bin\x264.exe" -Force
+        Remove-Item -Path "${FinalDir}\bin\zlib.def" -Force
+
+        # Copy native-compiled deps
+        Write-Step "Copy native-compiled $($Package.WinArch)/$($Package.Arch) files..."
```
PatTheMav, line 109: See above re. using array and a pipe.

CI/package-deps-windows.ps1
```diff
@@ -0,0 +1,159 @@
...
+    Ensure-Directory "${DepsBuildDir}\win32"
+    Ensure-Directory "${DepsBuildDir}\win64"
+    cd ..
+
+    $Packages = @(
+        [pscustomobject]@{Arch='x86'; WinArch='win32'}
+        [pscustomobject]@{Arch='x86_64'; WinArch='win64'}
+    )
+    Foreach ($Package in $Packages) {
+        if ($Package.Arch -eq "x86") {
```
PatTheMav, line 73: I've come to prefer this pattern:

```
    $QtDirectory = "${CheckoutDir}/../obs-build-dependencies/Qt_${WindowsQtVersion}/msvc2019$(if (${BuildArch} -eq "64-bit") { "_64" })"
    $DepsDirectory = "${CheckoutDir}/../obs-build-dependencies/dependencies${WindowsDepsVersion}/win$(if (${BuildArch} -eq "64-bit") { "64" } else { "32" })"
    $CefDirectory = "${CheckoutDir}/../obs-build-dependencies/cef_binary_${WindowsCefVersion}_windows$(if (${BuildArch} -eq "64-bit") { "64" } else { "32" })_minimal"
```
Matter of taste probably.

CI/package-deps-windows.ps1
```diff
@@ -0,0 +1,159 @@
...
+    tar -czf "${FileName}" -C "${DepsBuildDir}" *
+
+    Write-Info "All done!"
+}
+
+## MAIN SCRIPT FUNCTIONS ##
+function Print-Usage {
+    Write-Host "package-deps-windows.ps1 - Package script for ${ProductName}"
+    $Lines = @(
+        "Usage: ${MyInvocation.MyCommand.Name}",
```
PatTheMav, line 144: See above re. this not doing what we expect.

CI/windows/build_cmocka.ps1
```diff
@@ -0,0 +1,77 @@
...
+    cd "${DepsBuildDir}"
+
+    if ($Quiet) {
+        $CMAKE_OPTS = "-Wno-deprecated -Wno-dev --log-level=ERROR"
+    } else {
+        $CMAKE_OPTS = ""
+    }
+
+    Write-Step "Configure (${ARCH})..."
+    cmake -G "Visual Studio 16 2019" `
```
PatTheMav, line 33: I've come to prefer composing the CMake command this way:

```
   $CmakeCommand = @(
        "-S . -B `"${BuildDirectory}`"",
        "-G `"Visual Studio 16 2019`"",
        "-DCMAKE_GENERATOR_PLATFORM=`"${GeneratorPlatform}`"",
        "-DCMAKE_SYSTEM_VERSION=`"${CmakeSystemVersion}`"",
        "-DCMAKE_PREFIX_PATH=`"${CmakePrefixPath}`"",
        [OTHER_STUFF],
        "$(if (Test-Path Variable:$Quiet) { "-Wno-deprecated -Wno-dev --log-level=ERROR" })"
    )

    Invoke-Expression "cmake ${CmakeCommand}"
```

CI/windows/build_cmocka.ps1
```diff
@@ -0,0 +1,77 @@
...
+
+    Write-Step "Build (${ARCH})..."
+    cmake --build "cmocka_build\${CMAKE_BITNESS}" --config "${BuildConfiguration}"
+}
+
+function Install-Product {
+    cd "${DepsBuildDir}"
+
+    Write-Step "Install (${ARCH})..."
+    cmake --install "cmocka_build\${CMAKE_BITNESS}" --config "${BuildConfiguration}" --prefix "${DepsBuildDir}\${CMAKE_INSTALL_DIR}"
```
PatTheMav, line 49: Install prefix should've been set at configuration time via `-DCMAKE_INSTALL_PREFIX="[WHEREVER]"` already, but maybe matter of styling.

CI/windows/build_libopus.sh
```diff
@@ -0,0 +1,66 @@
...
+    cd "${PRODUCT_FOLDER}"
+
+    step "Configure ("${ARCH}")..."
+    git clean -dxf
+    ./autogen.sh
+    PKG_CONFIG_PATH="${BUILD_DIR}/lib/pkgconfig" \
+        LDFLAGS="-L${BUILD_DIR}/lib -static-libgcc" \
+        CPPFLAGS="-I${BUILD_DIR}/include -D_FORTIFY_SOURCE=0" \
+        ./configure \
+        --host=$WIN_CROSS_TOOL_PREFIX-w64-mingw32 \
```
PatTheMav, line 24: Use curlys (curls?): `${WIN_CROSS_TOOL_PREFIX}-w64-mingw32`

CI/windows/build_libvpx.sh
```diff
@@ -0,0 +1,97 @@
...
+#
+# This script file can be included in build scripts for Windows or run directly
+#
+################################################################################
+
+# Halt on errors
+set -eE
+
+_fixup_libs() {
+    vpxname=`find . -type f -iname libvpx*.dll`
```
PatTheMav, line 15: Use `UPPERCASE` for variables, use `$()` for subcommand invocations.

On second look I'd use `VPXNAME="$(find . -type f -iname "libvpx*.dll" | xargs basename)"`.

Be aware that `find` could yield multiple results, which `xargs` will work with, but your variable will contain multiple values. `head -n1` might be necessary then.

PatTheMav: This might be even better as it will invoke it for every file it finds:

```
find . -type f -iname "libvpx*.dll" | xargs -I{} sh -c "${WIN_CROSS_TOOL_PREFIX}-w64-mingw32-dlltool -m ${WIN_CROSS_MVAL} -d libvpx.def -l \"${BUILD_DIR}/bin/vpx.lib\" -D \"${BUILD_DIR}/bin/{}\""
```

CI/windows/patches/ffmpeg/FFmpeg-devel-avformat-hlsenc-remove-unnecessary-http-https-shutdown-status-operate.diff
```diff
@@ -0,0 +1,14 @@
+diff --git a/libavformat/hlsenc.c b/libavformat/hlsenc.c
```
PatTheMav, line 1: Maybe rename to `.patch` like the other files?

PatTheMav (changes requested):

.github/workflows/windows_deps.yml
RytoEX: I could, but the Bash was already written for the macOS workflow, and this job runs on Ubuntu, so I didn't care.

CI/build-deps-windows-native.ps1
```diff
@@ -0,0 +1,146 @@
...
+    $ErrorActionPreference = "Stop"
+
+    Build-Checks
+
+    if (!$CurrentDate) {
+        $CurrentDate = Get-Date -UFormat "%Y-%m-%d"
+    }
+    $FileName = "${ProductName}-win-native-${CurrentDate}-${CMAKE_INSTALL_DIR}.tar.gz"
+
+    if (!$SkipDependencyChecks) {
```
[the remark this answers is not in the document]
RytoEX: Yes, but I'd already written the vast majority of the script by the time we discussed that point, and I didn't want to delay this first iteration further.  I plan to change this later.

CI/include/build_support.sh
```diff
@@ -77,6 +90,33 @@ check_ccache() {
...
 
+check_git() {
+    step "Check git..."
+    if git --version >/dev/null 2>&1; then
+        GIT_VERSION="$(git --version | sed -e 's/git version //')"
+        info "Git version $GIT_VERSION available"
+
+        info "Check git config for user..."
+        git_user_email=$(git config --get user.email)
+        if [ -z "$git_user_email" ]; then
```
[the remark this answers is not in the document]
RytoEX: If you want to do `git cherry-pick`, yes.  See conversation in #78.

CI/include/build_support.sh
```diff
@@ -165,7 +206,12 @@ github_fetch() {
         fi
 
     else
-        git clone "https://github.com/${GH_USER}/${GH_REPO}.git" "$(pwd)"
+        if [ "${CI}" ] && [ "$(git_has_sparse_checkout)" ] && [ "${GIT_OPT_SPARSE}" ]; then
```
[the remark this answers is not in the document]
RytoEX: Here, I think I was trying to avoid a really long line, and I tend to lean towards breaking things into functions if they could conceivably be/become common functionality.

.github/workflows/windows_deps.yml
RytoEX: I could use PowerShell here, but since this runs on GHA rather than locally and I had the option of Bash as a shell _and_ this step was already written for the macOS jobs, I elected to leave it as is.  I didn't see a need to reinvent the wheel here.

.github/workflows/windows_deps.yml
```diff
@@ -0,0 +1,649 @@
...
+          else
+            echo "SEEKING_TESTERS=0" >> $GITHUB_ENV
+          fi
+
+          echo "CACHE_DATE=$(date +"%Y-%m-%d")" >> $GITHUB_ENV
+
+      - name: 'Setup build environment'
+        shell: bash
+        run: |
+          mkdir -p windows_native_build_temp/win${{ matrix.arch }}/bin
```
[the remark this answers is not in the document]
RytoEX: Same reasoning as above.  If these were in build scripts to be run on Windows, I would have used PowerShell, but since it's on GHA and Bash is available, I went with that, in part because fighting with Test-Path and New-Item can be a bit of a pain.

.github/workflows/windows_deps.yml
```diff
@@ -0,0 +1,649 @@
...
+      - name: 'Install dependency swig'
+        if: steps.swig-cache.outputs.cache-hit == 'true'
+        run: CI/windows/build_swig.ps1 -SkipDependencyChecks -BuildArch ${{ matrix.arch }}-bit -Install
+
+      - name: 'Package dependencies'
+        if: success()
+        working-directory: ${{ github.workspace }}/windows_native_build_temp/win${{ matrix.arch }}
+        shell: bash
+        run: |
+          CURRENT_DATE=$(date +"%Y-%m-%d")
```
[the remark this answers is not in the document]
RytoEX: Hmm.  If the concern is that it's a Windows Host, even on GHA, then I can sort of see the area of concern.  That said, GHA makes Bash available on Windows, so I'm not sure if this is something we _need_ to address right now.  I can make some notes to address this in a later update.

CI/build-deps-windows-cross.sh
```diff
@@ -0,0 +1,88 @@
...
+        PRODUCT_VERSION="${2}"
+        PRODUCT_HASH="${3}"
+
+        source "${CHECKOUT_DIR}/CI/windows/build_${1}.sh"
+    done
+
+    cd "${CHECKOUT_DIR}/windows/obs-dependencies-${ARCH}"
+
+    step "Cleanup unnecessary files..."
+    #find . \( -type f -or -type l \) \( -name "*.la" -or -name "*.a" \) | xargs rm
```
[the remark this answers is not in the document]
RytoEX: Done.  Will push update later.

CI/build-deps-windows-native.ps1
```diff
@@ -0,0 +1,146 @@
...
+    @('python', '3.6.2'),
+    @('vulkan', '1.2.131.2', '49d515f091d69c005a9e52e69829d28383fe157a764caebaa3dbc4b8d9bb383d'),
+    @('nasm', '2.15.01'),
+    @('swig', '3.0.12', '21ce6cbe297a56b697ef6e7e92a83e75ca41dedc87e48282ab444591986c35f5')
+)
+
+function Build-OBS-Deps-Main {
+    Ensure-Directory "${CheckoutDir}"
+    Write-Step "Fetching version tags..."
+    & git fetch origin --tags
```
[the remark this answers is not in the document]
RytoEX: I'm sort of leaning towards keeping it so that we can see the output in the GHA log to make sure that tags were indeed fetched.  Unless there is a concern that this may emit an error and therefore break the script?

CI/build-deps-windows-native.ps1
```diff
@@ -0,0 +1,146 @@
...
+    tar -czf "${FileName}" -C "${CMAKE_INSTALL_DIR}" *
+
+    Write-Info "All done!"
+}
+
+## MAIN SCRIPT FUNCTIONS ##
+function Print-Usage {
+    Write-Host "build-deps-windows-native.ps1 - Build script for ${ProductName}"
+    $Lines = @(
+        "Usage: ${MyInvocation.MyCommand.Name}",
```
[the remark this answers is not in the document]
RytoEX: Fixed.  Will push later.

CI/include/build_support.sh
```diff
@@ -57,9 +57,22 @@ BUILD_CONFIG="${BUILD_CONFIG:-RelWithDebInfo}"
 CI_WORKFLOW="${CHECKOUT_DIR}/.github/workflows/main.yml"
 CURRENT_ARCH="$(uname -m)"
 CURRENT_DATE="$(date +"%Y-%m-%d")"
+GIT_VERSION=""
 
 ## Utility functions ##
 
+is_gte() {
```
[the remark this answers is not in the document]
RytoEX: I'm inclined to leave this in-place in case we find another use for it somewhere else.

CI/include/build_support_windows.ps1
```diff
@@ -0,0 +1,579 @@
...
+
+    $Separators = "\/"
+    $SepArray = $Separators.ToCharArray()
+    return ${Path}.Substring(${Path}.LastIndexOfAny($SepArray) + 1)
+}
+
+function Get-UniquePath {
+    $OriginalPathString = $Env:Path
+    $OriginalPath = $OriginalPathString -split ';'
+    # Sort-Object -Unique is case-insensitive
```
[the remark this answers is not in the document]
RytoEX: That's not a line of code.  It's a comment about the behavior of `Sort-Object` vs. `Select-Object`.

CI/include/build_support_windows.ps1
```diff
@@ -0,0 +1,579 @@
...
+
+    $NewPath = $NewPathArray -join ';'
+
+    return $NewPath
+}
+
+function Safe-Fetch {
+    Param(
+        [Switch] $UseCurl,
+        [Switch] $CurlNoContinue,
```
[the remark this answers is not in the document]
RytoEX: Yeah, I left in a lot of `UPPER_CASE` variable names due to old porting from the existing support include scripts and I didn't want to inadvertently break things by fixing the variable names to be more PowerShell-esque like `CapitalCase`.  Though, doing so is in my local notes for future fixes and polish.

CI/include/build_support.sh
```diff
@@ -77,6 +90,33 @@ check_ccache() {
     fi
 }
 
+check_git() {
+    step "Check git..."
+    if git --version >/dev/null 2>&1; then
+        GIT_VERSION="$(git --version | sed -e 's/git version //')"
+        info "Git version $GIT_VERSION available"
```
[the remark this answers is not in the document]
RytoEX: Fixed.  Will push later.

CI/include/build_support_windows.ps1
```diff
@@ -0,0 +1,579 @@
...
+}
+
+function Check-Curl {
+    if (!(Test-CommandExists "curl.exe") -and !$NoChoco) {
+        Write-Step "Install curl from chocolatey..."
+        Invoke-Expression "choco install -y curl"
+    }
+
+    # TODO: implement a way to force using/installing curl.exe from chocolatey
+    #$CURLCMD = "C:\ProgramData\chocolatey\bin\curl.exe"
```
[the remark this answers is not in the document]
RytoEX: Hmm.  I'd actually left this in as a reminder of where Chocolatey installs curl, but I see that because it's originally a line of code its purpose is confusing.  I might leave the path in place, but try to make it more part of the TODO entry rather than a pure line of code simply commented out.

CI/include/build_support_windows.ps1
```diff
@@ -0,0 +1,579 @@
...
+        # TODO: patch failed in my tests, hence using git apply above
+        patch -g 0 -f -p1 -i "${PATCH_FILE}"
+    }
+}
+
+function Check-Archs {
+    Write-Step "Check Architecture..."
+
+    if ("${BuildArch}" -eq "64-bit") {
+        $script:ARCH = "x86_64"
```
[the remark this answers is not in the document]
RytoEX: I'd prefer the same (CapitalCase).  It's on my list of polish changes to make in a later iteration.

CI/include/build_support_windows_cross.sh
```diff
@@ -0,0 +1,193 @@
...
+        esac
+    done
+}
+
+_build_checks() {
+    PRODUCT_NAME_U="$(echo ${PRODUCT_NAME} | tr [a-z] [A-Z])"
+    CI_PRODUCT_VERSION=$(/bin/cat "${CI_WORKFLOW}" | /bin/sed '/windows-deps-build-cross-compile/,/defaults:/!d' | /bin/sed -En "s/[ ]+${PRODUCT_NAME_U}_VERSION: '(.+)'/\1/p")
+    CI_PRODUCT_HASH=$(/bin/cat "${CI_WORKFLOW}" | /bin/sed '/windows-deps-build-cross-compile/,/defaults:/!d' | /bin/sed -En "s/[ ]+${PRODUCT_NAME_U}_HASH: '([0-9a-f]+)'/\1/p")
+
+
```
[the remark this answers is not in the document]
RytoEX: Fixed.

CI/package-deps-windows.ps1
```diff
@@ -0,0 +1,159 @@
...
+    } elseif (!(Test-Path "${NativeDirBase}\win32" -PathType "Container")) {
+        Caught-Error "Missing native build in ${NativeDirBase}\win32"
+    } elseif (!(Test-Path "${NativeDirBase}\win64" -PathType "Container")) {
+        Caught-Error "Missing native build in ${NativeDirBase}\win64"
+    }
+
+    Remove-ItemIfExists "${DepsBuildDir}"
+    Ensure-Directory "${DepsBuildDir}\win32"
+    Ensure-Directory "${DepsBuildDir}\win64"
+    cd ..
```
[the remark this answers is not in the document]
RytoEX: In this case, I'm changing back to "${DepsBuildDir}", and I believe that's a full path, so I could certainly just use that directly since it is already stored as a variable, and I think that's the simplest change to make here.  I think I was being lazy because I know that I've only just created a directory one level down, and in this case it indicates the hierarchy/location clearly enough since it's a single traversal (one `..`) rather than several.

I'm not sure about whether or not it's safer.  I'm pretty sure an immediate parent directory must exist for any path except root.  Could you perhaps elaborate?

CI/package-deps-windows.ps1
RytoEX: More PowerShell-esque, perhaps.  Simpler?  Debatable.

For now, I'm content to leave it as-is since it works and leave this question to a future change.

CI/package-deps-windows.ps1
RytoEX: Similar to above, I'll probably put that on a list to investigate for a future cleanup effort.

CI/package-deps-windows.ps1
RytoEX: I find that inline `if` or ternary statements are a bit jarring or unclear.  Admittedly, in such cases, I prefer using an intermediate variable.

CI/windows/build_cmocka.ps1
RytoEX: Hmm.  I'll mark this down as "PowerShell-isms to investigate/implement later".

CI/windows/build_cmocka.ps1
RytoEX: I believe this is where I picked up this pattern:  https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-2.html

CI/windows/build_libopus.sh
RytoEX: This is probably safest?  It works right now though, so I might push this off till later.
```
"${WIN_CROSS_TOOL_PREFIX}-w64-mingw32"
```

CI/windows/build_libvpx.sh
RytoEX: > Be aware that `find` could yield multiple results

In a clean build directory, there _should_ be only one result for this.  The only reason for the wildcard is to account for the version suffix (in this case, "-1").  We could probably detect what the suffix should be (e.g., from `build/make/version.sh` or the git tag version), and therefore the full output filename, instead of messing with find.  I just didn't do that for this implementation because what I had at the moment worked.

> `head -n1` might be necessary then.

If there were multiple results, would this return the latest/newest libvpx?  Would that be desirable?

CI/windows/patches/ffmpeg/FFmpeg-devel-avformat-hlsenc-remove-unnecessary-http-https-shutdown-status-operate.diff
RytoEX: Ah, I think the reason this is a .diff is because that's the extension that it came with from the FFmpeg patch system, and I was hesitant/reluctant to change it.  Thoughts?
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20210913021204.22138-1-lq@chinaffmpeg.org/

RytoEX:
I've addressed the feedback items that I've deemed critical or that were trivial enough to implement without concerns.  I've also updated the original post with some additional information, particularly caveats.

@DDRBoxman @PatTheMav 
Ideally, I'd like to get this, and its dependent/included PRs, reviewed and merged as soon as possible so that we can tag a Windows deps release and begin iterating from there.  There is some additional cleanup and polish that could probably be done in this PR, and indeed, I plan to do so at some point and am maintaining a to-do list locally.  However, I'm content to make such minor changes in future PRs.  For now, I'd like us to focus only on critical issues for these PRs.

@WizardCM has been helping me debug and test these build scripts for the past few weeks.  Hopefully he can provide insight, approval, or some additional testing of the build artifacts.

CI/include/build_support.sh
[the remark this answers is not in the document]
PatTheMav: I'd rather follow YAGNI here, but if we want to make this multi-purpose, I'd rather have a `check_version` function that takes 3 arguments (1. value, comparator, 2. value) and implement the different comparisons there (equality/non-equality is simple, whereas `gte` and `lte`  just change a single parameter to `sort` or changes `head` for `tail`).

Also use return codes instead of strings, that's very "unshell"-ish.

CI/package-deps-windows.ps1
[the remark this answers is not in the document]
PatTheMav: In that particular case it should be fine, I've been bitten by functions/calls that change the working directory and so the state upon returning was not as expected.

CI/package-deps-windows.ps1
PatTheMav: My idea was that adding additional entries to remove is simpler via an additional entry in a array instead of adding a whole function call.

CI/windows/build_cmocka.ps1
PatTheMav: Yeah it's canonical, the parameter is there for a reason; my expectation would've been that setting that prior to running `install` avoids another CMake configuration step (as it doesn't have to generate all the `cmake_install.cmake` files, but that's just an optimisation.

CI/windows/build_libvpx.sh
PatTheMav: Not the newest, just the first. You'd probably have to adjust the find parameters, but you can optimise this later.

CI/windows/patches/ffmpeg/FFmpeg-devel-avformat-hlsenc-remove-unnecessary-http-https-shutdown-status-operate.diff
PatTheMav: Happened to me as well, I renamed it to patch for consistency with our repos.

CI/package-deps-windows.ps1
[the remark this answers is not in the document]
RytoEX: Ah, perhaps you mean that the working directory is changed _in general_ away from where the user may have started?  In that case, yes, the script should be a nice player and return the shell's working directory to where the user started.  As this is a pretty minor change and it's in the packaging script, I may make that now instead of pushing it till later.

CI/include/build_support.sh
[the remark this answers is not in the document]
RytoEX: For a future iteration, I think that would be a good change.  For this iteration, I wrote what I needed without building a full set of comparison functions/operations.

Either way, I'll mull this over for a bit for:
1. whether this remains a separate function or if I merge it into `github_fetch`, though I'd probably lean towards the former
2. whether this will be implemented now or pushed till a later iteration (leaning towards the latter)

CI/package-deps-windows.ps1
RytoEX: > My idea was that adding additional entries to remove is simpler via an additional entry in a array instead of adding a whole function call.

That particular aspect may be simpler, yes.  I'm fine with that in general, just perhaps for a future iteration unless I get to it very quickly.

CI/windows/build_cmocka.ps1
RytoEX: I'm honestly not sure what the best practice here is.  I have a hazy recollection that doing it this way avoided some issue, but I may be misremembering.  I'll make a note to look into this further.

CI/windows/patches/ffmpeg/FFmpeg-devel-avformat-hlsenc-remove-unnecessary-http-https-shutdown-status-operate.diff
RytoEX: Oh.  It occurs to me that I'm not actually using that patch here at all, because we're still on FFmpeg 4.2.4 here, where this patch is not needed.  I can probably remove it since it's not OS-specific, and you've already added it to the non-OS-specific patches directory.

RytoEX:
Dependent PRs have been merged.  CI is green.  Will be merging this shortly.

CI/windows/patches/ffmpeg/FFmpeg-devel-avformat-hlsenc-remove-unnecessary-http-https-shutdown-status-operate.diff
RytoEX: Removed this.

PR #81, opened 2021-11-15
Fix github_fetch argument count check

RytoEX:
### Description

To add support for git sparse-checkout, I had tried to change github_fetch to allow it to accept three OR four arguments. Instead of causing it to error out if less than three arguments were supplied, I'd caused it to error out if three or less arguments weere supplied. This is because I'd used the -le operator (less than or equal to) instead of the -lt operator (less than). Use the correct -le operator here. 

### Motivation and Context

Fix builds.

### How Has This Been Tested?

Tested the comparison operator change in local test scripts.

### Types of changes

### Checklist:

### DIFF
--- a/CI/include/build_support.sh
+++ b/CI/include/build_support.sh
@@ -189,7 +189,7 @@ check_and_fetch() {
 }
 
 github_fetch() {
-    if [ $# -le 3 ]; then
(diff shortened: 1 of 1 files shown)


RytoEX:
CI is green, so I'm merging this fix.

PR #82, opened 2021-11-18
Add the AJA NTV2 lib

DDRBoxman:
### Description

This leverages the cmake install changes recently made on the Aja repo to let us easily build and package the Aja lib.

### Checklist:

- [ ] All commit messages are properly formatted and commits squashed where appropriate.


DDRBoxman:
We actually end up with a static library for this one (.a) so we can't / don't need install_name_tool

PatTheMav:
It still forces a macOS 10.12 deployment target _and_ uses code that has been deprecated since 10.12. I'm not really happy merging code that ignored deprecation warnings for 10 years now, especially with a growing emphasis on M1 (which uses macOS 11 as a baseline).

At _least_ I'd require us to have a patch file that fixes that CMake module that allows us to specify a proper deployment target per architecture instead of forcing 10.12.

PatTheMav:
So I checked out the project itself and I don't think we need to jump through all those annoying hoops - just patch the project with this diff to fix some remaining issues with their CMake files, and it will build a shared lib that installs into the desired install prefix and doesn't require any fixups (because CMake does the fixups itself when you `install` dylibs).

```
diff --git a/cmake/BuildOptions.cmake b/cmake/BuildOptions.cmake
index 7de78d4..64842d5 100644
--- a/cmake/BuildOptions.cmake
+++ b/cmake/BuildOptions.cmake
@@ -5,7 +5,9 @@ set(CMAKE_DEBUG_POSTFIX d)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 # macOS build settings
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "Minimum OS X deployment version" FORCE)
+if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.12" CACHE STRING "Minimum OS X deployment version")
+endif()
 # TODO(paulh): generate universal builds up front
 # set(CMAKE_OSX_ARCHITECTURES arm64 x86_64)
 
diff --git a/ajalibraries/ajantv2/CMakeLists.txt b/ajalibraries/ajantv2/CMakeLists.txt
index 5837dc1..a382a6f 100644
--- a/ajalibraries/ajantv2/CMakeLists.txt
+++ b/ajalibraries/ajantv2/CMakeLists.txt
@@ -484,6 +484,13 @@ if (AJA_BUILD_SHARED)
     target_include_directories(ajantv2shared PUBLIC ${AJANTV2_TARGET_INCLUDE_DIRS})
     target_link_libraries(ajantv2shared PUBLIC ${EXTRA_LIBRARIES} ${EXTRA_DYNAMIC_LIBRARIES})
     target_compile_definitions(ajantv2shared PUBLIC ${EXTRA_DYNAMIC_DEFINES})
+
+    install(TARGETS ajantv2shared
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+        FRAMEWORK DESTINATION ${CMAKE_INSTALL_PREFIX}
+        PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ajalibraries/ajantv2)
 endif()
 
 # cmake install command
```

You want to run CMake as:

`cmake -S . -B build -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13 -DCMAKE_OSX_ARCHITECTURES=arm64 -G Ninja -DAJA_BUILD_OPENSOURCE=ON -DAJA_BUILD_SHARED=ON -DCMAKE_INSTALL_PREFIX=/Users/patthemav/GitHub/ntv2/build/install -DCMAKE_CXX_COMPILER_LAUNCHER=ccache`

PR #96, opened 2022-02-20
Pin luajit 2.1 mirror=

waiki510:
### Checklist:

- [ ] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ ] My code is not on the master branch.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.


waiki510:
THANK U

RytoEX:
This appears to be based off of a very old commit from a branch that is no longer in sync with the current build system.  We already use LuaJIT 2.1 for macOS.  We do not currently use LuaJIT 2.1 on Windows because we weren't prepared to make that update with the most recent OBS Studio release.

I'll address this in a PR when we're ready to make that change.

PR #100, opened 2022-03-22
Update ntv2 hash and cmake command for recent AJA NTV2 CMake changes

paulh-aja:
### Description

This change updates the NTV2 hashes to the latest NTV2 16.2 bugfix2 release commit ([6c44a9476e6f3792671c147a28d103ce2adf8d6e](https://github.com/aja-video/ntv2/commit/6c44a9476e6f3792671c147a28d103ce2adf8d6e)), and tweaks the CMake commands to accommodate the latest state of the NTV2 CMake scripts.

### Motivation and Context

Update to the latest AJA NTV2 16.2 release code and CMake.

### How Has This Been Tested?

Tested the Windows `build_ntv2.ps1` script with the updated NTV2 hash and verified that the files are building and output to the correct directory structure (`windows_native_build_temp/win64/include` and `lib`) within the obs-deps project.

I do not have a way to test the macOS changes right now but I've made what I believe to be the correct changes here.

### Types of changes

### Checklist:


paulh-aja:
The CI checks look to have worked properly. It looks like the NTV2 files were placed in the correct output destination by cmake install.

tytan652:
Is `-DAJA_INSTALL_SOURCES=ON` really required ?

RytoEX (approved):
Looks fine to me. If @PatTheMav or any other macOS-based dev approves the macOS changes, then this can be merged.

RytoEX:
> Is `-DAJA_INSTALL_SOURCES=ON` really required ?

I did miss this question while reviewing. Could you please answer this, @paulh-aja ?

CI/macos/build_ntv2.sh
```diff
@@ -16,10 +16,14 @@ _build_product() {
 
     step "Configure ("${ARCH}")..."
     cmake -S . -B build_${ARCH} -G Ninja ${CMAKE_CCACHE_OPTIONS} \
-        -DCMAKE_INSTALL_PREFIX="${BUILD_DIR}" \
+        -DCMAKE_INSTALL_PREFIX="${BUILD_DIR}\include" \
```
PatTheMav, line 19: You should not specify the `include` directory as a prefix. The prefix constitutes the "base" directory for installation, which you extend inside your CMake script by specifying either a custom subdirectory (`${CMAKE_INSTALL_PREFIX}/<YOUR_CUSTOM_DIR>`), or - which is preferred - the common directories set by `include(GNUInstallDirs)` (`${CMAKE_INSTALL_INCLUDEDIR}`, `${CMAKE_INSTALL_LIBDIR}`) which will be automatically composed based off the prefix directory.

PatTheMav: So the common approach is:

1. Set the prefix
2. Tell CMake the directory name(!) within the path specified by the prefix 

So if you run
```
install(TARGETS <YOUR_TARGET>
  RUNTIME DESTINATION bin
  LIBRARY DESTINATION lib
  ARCHIVE DESTINATION lib
  PUBLIC_HEADER DESTINATION include)
``` 
CMake will automatically install headers in `<PREFIX>/include`, libraries in `<PREFIX>/lib`, etc.

handledexception:
> > Is `-DAJA_INSTALL_SOURCES=ON` really required ?
> 
> I did miss this question while reviewing. Could you please answer this, @paulh-aja ?
It is required to set both AJA_INSTALL_SOURCES and AJA_INSTALL_HEADERS to copy the .h and .cpp files into the CMake install directory. We added these variables to keep the installation of sources and headers separate so that we could more easily produce "header-only" distributions of NTV2. We could certainly look at changing how we go about this though if it is not a good practice in CMake to separate things in this way.

tytan652:
> It is required to set both AJA_INSTALL_SOURCES and AJA_INSTALL_HEADERS to copy the .h and .cpp files into the CMake install directory. We added these variables to keep the installation of sources and headers separate so that we could more easily produce "header-only" distributions of NTV2. We could certainly look at changing how we go about this though if it is not a good practice in CMake to separate things in this way.

I'm going ask more precise:
Is `-DAJA_INSTALL_SOURCES=ON` really required ? I don't think we need those `.cpp` files to be added inside obs-deps archive, except if it's really required to link against the lib which shall not be the case.

handledexception:
> > It is required to set both AJA_INSTALL_SOURCES and AJA_INSTALL_HEADERS to copy the .h and .cpp files into the CMake install directory. We added these variables to keep the installation of sources and headers separate so that we could more easily produce "header-only" distributions of NTV2. We could certainly look at changing how we go about this though if it is not a good practice in CMake to separate things in this way.
> 
> I'm going to ask more precisely: Is `-DAJA_INSTALL_SOURCES=ON` really required ? I don't think we need those `.cpp` files to be added inside obs-deps archive, except if it's really required to link against the lib which shall not be the case.

My understanding is that the OBS CI builds for macOS and Windows are building the AJA NTV2 SDK static lib from sources. If that is still the case then the sources should go into the CMake install directory. Maybe I'm missing something here. @DDRBoxman do I still have my facts straight?

PatTheMav:
You usually don't need the source files to link a static library, the public header files with the necessary public symbols should suffice.

handledexception:
> You usually don't need the source files to link a static library, the public header files with the necessary public symbols should suffice.
Right, but I thought OBS CI was building the entire ajantv2 lib itself from sources. Is that not the case?

tytan652:
obs-deps is meant to create archive/tarball of compiled deps for OBS for Windows ans macOS. Like that OBS CI only build OBS Studio itself and link it to those compiled deps.

RytoEX:
I think we're still waiting on the resolution of the two conversations above regarding [`CMAKE_INSTALL_PREFIX`](https://github.com/obsproject/obs-deps/pull/100#discussion_r835917152) and `AJA_INSTALL_SOURCES` before landing this.

To reiterate: the obs-studio CI does not build the AJA libraries. We build those here, on the obs-deps CI. Then, the obs-studio CI downloads the pre-built libraries and uses those to build the AJA plugins (aja and aja-output-ui). The obs-deps package should not require .cpp files to be installed/copied to the final package, only header files (.h/.hpp). For examples of this final result, check the resulting `./include` folder of the current obs-deps releases.

Please let us know if you have any questions.

paulh-aja:
I've removed the AJA_INSTALL_SOURCES variable setters. I am going to add one more CMake change into the NTV2 repo this weekend to resolve the CMAKE_INSTALL_PREFIX discussion.

PatTheMav:
@paulh-aja Feel free to ping me in a comment once you force-pushed the updates so we can get this merged as soon as possible.

RytoEX:
@paulh-aja Any updates on this?

paulh-aja:
> @paulh-aja Any updates on this?

I've pushed the changes to our NTV2 repo but I still need to update this PR with the updated changes. I'll try to get that completed this week.

tytan652:
16.2-bugfix3 still put `ajadriver` and `ajalibraries` headers folder in `CMAKE_INSTALL_PREFIX` and not in the include directory.

It seems when installing headers `CMAKE_INSTALL_PREFIX` is used rather than `CMAKE_INSTALL_INCLUDEDIR`:
- https://github.com/aja-video/ntv2/blob/v16.2-bugfix3/ajadriver/CMakeLists.txt#L171
- https://github.com/aja-video/ntv2/blob/v16.2-bugfix3/ajalibraries/ajantv2/CMakeLists.txt#L176
- https://github.com/aja-video/ntv2/blob/v16.2-bugfix3/ajalibraries/ajantv2/CMakeLists.txt#L527

paulh-aja:
OK, I think I understand this a bit more now.

The way we are currently using CMake Install internally within the NTV2 repo is to assemble a source/header/binary distribution of the NTV2 SDK. The sources/headers which contribute to the library build are deployed by `cmake --install` into the `CMAKE_INSTALL_PREFIX` directory, under the same tree structure as the original `ajalibraries` tree.

If I change the base install directory for the headers to use `CMAKE_INSTALL_INCLUDEDIR` instead of `CMAKE_INSTALL_PREFIX`, there doesn't seem to be an analogous default directory for placing source files (.c/.cpp, etc). This essentially means in order to fix this as requested for the obs-deps project, it will break our existing CI builds and force me to have to maintain two paths, one for headers and one for sources.

Is the request to use the `CMAKE_INSTALL_INCLUDEDIR` path as the destination for installing headers a matter of "best practices" in the case of the obs-deps build? Perhaps we are not using CMake install "correctly" in our internal CI builds to assemble a source distribution in this way?

tytan652:
Since [ntv2](https://github.com/aja-video/ntv2) seem to no have issue tracker, I'm going to comment about bugfix4 here:
- https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L63
  - I don't know if all those cmake file are required to be installed when we use/install only the library.
  - https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80, this create a `CMakeLists.txt` folder.
  
If those are not problematic for obs-deps, bugfix4 is fine.

paulh-aja:
> Since [ntv2](https://github.com/aja-video/ntv2) seem to not have issue tracker, I'm going to comment about bugfix4 here:
> 
> * https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L63
>   
>   * I don't know if all those cmake file are required to be installed when we use/install only the library on only one type of OS.
>   * https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80, this create a `CMakeLists.txt` folder.
> 
> If those are not problematic for obs-deps, bugfix4 is fine.

Unfortunately we don't have the issue tracker set up yet. I need to talk to my manager and IT about enabling it now that we're starting to have community interaction with the repo.

The CMake files are not required to be installed for the lib. I pushed the latest state of our CMakeLists from our internal NTV2 repo. As I mentioned our CMake Install directives are also intended for assembling an NTV2 distribution from the repo, which is what those install directives are for. I may need to put some more thought into how we expose that functionality as we go along.

I will take another look at https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80.

I've updated this PR to point to the bugfix4 commit hash.

tytan652:
> I will take another look at https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80.

I just realise it now, but the whole install block create unneeded subfolders, removing the filename in the destination part should fix it.

paulh-aja:
OK I am going to fix [CMakeLists.txt](https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80.) and re-make the bugfix4 release.

paulh-aja:
Updated the NTV2 hashes to point to "bugfix5" release.

paulh-aja:
> > I will take another look at https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80.
> 
> I just realise it now, but the whole `install` block creates unneeded subfolders, removing the filename in the destination part should fix it.
> 
> Edit: The block is from line 64 to 80.

Ah...dang you're right. I'll have to fix the whole block. Arg... OK back to the drawing board.

RytoEX:
> > > I will take another look at https://github.com/aja-video/ntv2/blob/v16.2-bugfix4/CMakeLists.txt#L80.
> > 
> > 
> > I just realise it now, but the whole `install` block creates unneeded subfolders, removing the filename in the destination part should fix it.
> > Edit: The block is from line 64 to 80.
> 
> Ah...dang you're right. I'll have to fix the whole block. Arg... OK back to the drawing board.
> 
> Edit: Pushed a fix and updated the tag v16.2-bugfix5 to point to the new commit [aja-video/ntv2@c1f708c](https://github.com/aja-video/ntv2/commit/c1f708cc01ad1156e71085a3b9cf44f033fe39cd). I've updated this obs-deps PR again to point to this new NTV2 hash.

Edits don't send notification emails, so I missed this edit. Does that mean this is ready for re-review?

PatTheMav:
@paulh-aja I think there are no further open review notes by us - feel free to undraft the PR. We'd really like to get this merged.

tytan652:
`AJA_INSTALL_SOURCES` is `ON` by default please just add `-DAJA_INSTALL_SOURCES=OFF` to the PR and undraft the PR.

https://github.com/aja-video/ntv2/blob/v16.2-bugfix5/cmake/AJABuildOptions.cmake#L39

PatTheMav:
@paulh-aja I did a test of this today and even if `AJA_INSTALL_SOURCES` set to `OFF`, there's still stray files being installed into the install destination:

* The main `CMakeLists.txt` is installed into the prefix root directory
* The entire `cmake` directory is installed into the prefix root directory

paulh-aja:
> @paulh-aja I did a test of this today and even if `AJA_INSTALL_SOURCES` set to `OFF`, there's still stray files being installed into the install destination:
> 
> * The main `CMakeLists.txt` is installed into the prefix root directory
> * The entire `cmake` directory is installed into the prefix root directory

Ah, I fixed that in the NTV2 CMakeLists but forgot to spin a new release tag. I will do that an update this PR again.

paulh-aja:
Updated the PR again with a new NTV2 commit hash. The NTV2 commit sets AJA_INSTALL_SOURCES to OFF by default.

tytan652:
You forgot to put the PR ready for review.

paulh-aja:
When this PR is merged into obs-deps, this plugin will need to be merged into OBS: https://github.com/obsproject/obs-studio/pull/6194

RytoEX (approved):
The updates seem to achieve what we wanted. Re-approving.

RytoEX:
Merging shortly. Note that this will still need a new release before the obs-studio CI starts using it.

paulh-aja:
Are these deps being built and used in the obs-studio CI builds now? If everything on the OBS side is still building correctly then we may not need https://github.com/obsproject/obs-studio/pull/6194 anymore. If OBS hasn't yet been updated to build with deps from this PR then I will need to do a test and make sure OBS can still build following this change, which updates the obs-deps to use the NTV2 16.2 SDK version.

RytoEX:
Per [my comment](https://github.com/obsproject/obs-studio/pull/6194#issuecomment-1144088298) on obsproject/obs-studio#6194, we switched to using the 2022-05-23 obs-deps as of https://github.com/obsproject/obs-studio/pull/6576.

PR #102, opened 2022-04-10
macos/patches: Backport QT macOS docking fixes

gxalpha:
### Description

Backports qt/qtbase@78b6050 and qt/qtbase@3224c6d, fixing the broken macOS
docking and undocking.

These fixes are documented to fix [QTBUG-70137](https://bugreports.qt.io/browse/QTBUG-70137), where docks will freak out when getting dragged out of the UI; however they also happens to fix [QTBUG-64483](https://bugreports.qt.io/browse/QTBUG-64483) (floating docks randomly deciding they lost their will to live, and no longer being re-dockable) (@RytoEX has been watching that one).

These two bugs were the biggest troublemakers in terms of docking and undocking on macOS.
The patches are no longer needed once we move to QT 6.2 (which is where they are from), but for now I think they are really important (and perhaps could even justify a macOS hotfix on their own).

### Motivation and Context

Docking can get very broken on macOS.
This makes it better. Not perfect, but much better..

### How Has This Been Tested?

Built on GHA and compiled my OBS against it.
Test QT artifacts [here](https://github.com/gxalpha/obs-deps/actions/runs/2143386691), so that you don't need to wait two hours for QT to compile.

### Types of changes

### Checklist:

### DIFF
--- a/CI/macos/build_qt.sh
+++ b/CI/macos/build_qt.sh
@@ -17,6 +17,9 @@ _patch_product() {
     if [ -z "${SKIP_UNPACK}" ]; then
         step "Apply patches..."
         apply_patch "${CHECKOUT_DIR}/CI/patches/QTBUG-74606.patch" "6ba73e94301505214b85e6014db23b042ae908f2439f0c18214e92644a356638"
+        apply_patch "${CHECKOUT_DIR}/CI/macos/patches/QTBUG-70137-1.patch" "216be72245a80b7762dc2e2bd720a4ea9b9c423ce9d006cce3985b63c0269ba3"
+        apply_patch "${CHECKOUT_DIR}/CI/macos/patches/QTBUG-70137-2.patch" "92d49352c321c653d6f5377e64603e48b38a9c1ec87a8956acba42459c151e42"
+        apply_patch "${CHECKOUT_DIR}/CI/macos/patches/QTBUG-70137-3.patch" "f8b220a444fcd0e121b8643e7526af33a4f30e0c85d11c28d40fcc7072d56783"
         apply_patch "${CHECKOUT_DIR}/CI/macos/patches/QTBUG-88495.patch" "d60d663d2d940aa21cbcda65b1e60c4ecb1ec1900736e896367e5436d121206e"
         apply_patch "${CHECKOUT_DIR}/CI/macos/patches/QTBUG-97855.patch" "d8620262ad3f689fdfe6b6e277ddfdd3594db3de9dbc65810a871f142faa9966"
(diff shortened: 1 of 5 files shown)


RytoEX:
Do we not need the third patch that amends the second patch? See:
* https://github.com/qt/qtbase/commit/ba3e1fe09b7d921985e21d857a1d566465095e69
* https://codereview.qt-project.org/c/qt/qtbase/+/362850

RytoEX (changes requested):
Commit message nit: QT -> Qt

gxalpha:
Fixed the commit message, I'll leave the decision on https://github.com/qt/qtbase/commit/ba3e1fe09b7d921985e21d857a1d566465095e69 to you, see my comment above.

RytoEX:
We can leave the third patch out if it's truly not needed. We can always land this now and add it later if we find a need. I'm honestly not entirely sure what cases/issues it fixes.

On the other hand, it might be less annoying to just add it now so we don't have to re-release later should we find it useful.

Thoughts @PatTheMav ?

PatTheMav:
There is the potential of the omitted patch fixing other behaviour that later patches somehow rely on, but until that is confirmed I'm fine with it.

RytoEX:
> We don't set a custom title bar widget anywhere

We don't currently, but see https://github.com/obsproject/obs-studio/pull/6358. I don't know if that will be merged, but I'm tempted to err on the side of caution and just include the third patch if that PR would be affected. I'm leaning towards just including the extra patch since it was tagged on the QTBUG, and I don't want to have to remember that we didn't include it if this somehow later comes up.

Otherwise, this seems fine to me, so it will likely be merged one way or another.

gxalpha:
I also saw that PR, and with that in mind I also lean towards including the patch - I'll add it to this PR tomorrow or so

gxalpha:
Added the third patch.
Didn't test it locally since building Qt takes ages on my machine, but as long the patch applies cleanly on CI this should be good (can't see the patch doing any harm).

RytoEX:
> Added the third patch. Didn't test it locally since building Qt takes ages on my machine, but as long the patch applies cleanly on CI this should be good (can't see the patch doing any harm tbh).
> 
> Edit: CI restored from cache. Are you kidding me? :D

Yeah, the cache key is `'qt-cache-${{ matrix.arch }}-${{ env.QT_REVISION }}'`, so any build within 7 days that doesn't change the key would use the cache and then skip building. It's kind of a shame it won't restore the cache and then check for new patches. We could probably do it, but it would require reworking the skip step logic.

RytoEX (approved):

PR #104, opened 2022-04-27
CI: Optimize macOS Qt build

RytoEX:
### Description

Port some configuration options from our Windows Qt build to our macOS Qt build to reduce the build time and build size.

### Motivation and Context

Reducing the build time helps us validate PRs and commits faster and lets us free up macOS CI instances faster.

### How Has This Been Tested?

I let this build on my fork's CI. I checked the build time and artifact size. Both build time and artifact size were reduced from before these changes.

If we want to be thorough, someone should build OBS on macOS against this Qt build to make sure that nothing breaks.

#### Before PR
image
##### Build time (build only)
* x86_64: 1h09m35s
* arm64: 1h36m29s

##### Build size (compressed CI artifact)
* x86_64: 18.9 MB
* arm64: 20.5 MB
* universal: 28.3 MB

##### Build size (uncompressed)
* x86_64: 83.2 MB
* arm64: 99.9 MB
* universal: 140 MB

#### After PR

image

##### Build time (build only)
* x86_64: 43m15s
* arm64: 1h13m13s

##### Build size (compressed CI artifact)
* x86_64: 12.8 MB
* arm64: 15.3 MB
* universal: 19.2 MB

##### Build size (uncompressed)
* x86_64: 62.0 MB
* arm64: 78.1 MB
* universal: 100 MB

That's roughly 54% of the original build time (~46% reduction) for x86_64 and 76% for arm64 (24% reduction). Build sizes have been reduced to between roughly 68% and 78% of their original sizes (about 22% to 32% reductions).

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -397,7 +397,7 @@ jobs:
         arch: ['x86_64', 'arm64']
     env:
       MACOSX_DEPLOYMENT_TARGET: '10.13'
-      QT_REVISION: '03'
+      QT_REVISION: '04'
       QT_VERSION: '5.15.2'
       QT_HASH: '3a530d1b243b5dec00bc54937455471aaa3e56849d2593edb8ded07228202240'
       BLOCKED_FORMULAS: 'llvm gcc postgresql openjdk sox libsndfile flac libvorbis opusfile libogg composer php gd freetype fontconfig webp libpng lame libtiff opus kotlin sbt libxft libxcb'
--- a/CI/macos/build_qt.sh
+++ b/CI/macos/build_qt.sh
@@ -104,13 +104,22 @@ _configure_qt_x86_64() {
     ../configure ${CMAKE_CCACHE_OPTIONS:+-ccache} ${QMAKE_QUIET:+-silent}  \
         -release -opensource -confirm-license -system-zlib \
         -qt-libpng -qt-libjpeg -qt-freetype -qt-pcre \
-        -nomake examples -nomake tests -no-glib \
+        -nomake examples -nomake tests \
+        -no-compile-examples -no-dbus -no-glib \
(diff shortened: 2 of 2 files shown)


derrod:
Just out of curiousity does it actually try to build `qtwayland` on macOS?

RytoEX:
> Just out of curiousity does it actually try to build `qtwayland` on macOS?

It was already being skipped before this change, so I'm not sure.

RytoEX:
I noticed that we explicitly selected `-dbus-runtime`, so I did not drop it in this PR. We could additionally skip Qt D-Bus if that's not being used for anything. We currently disable it on Windows, but I'm not sure if it's used for something in macOS builds. It currently adds a couple of minutes to the build time.

gxalpha (approved):
OBS builds and runs successfully with artifacts of this PR

RytoEX:
> OBS builds and runs successfully with artifacts of this PR

Glad to hear it.

@PatTheMav Any comments?

PatTheMav:
Changes themselves look fine, if @gxalpha says OBS compiles fine with the generated Qt artefact, that's good enough for me.

Regarding dbus, I don't really know how that might be interconnected with other modules, but afaik macOS doesn't use dbus.

PatTheMav (approved):

RytoEX:
I've removed dbus from the build and managed to shave off about two minutes from the x86_64 build time. The arm64 build took about 3 minutes longer, but I think that's just due to CI variance, and that the average time over the long run will be shorter. The Qt build artifacts are slightly smaller now:
* x86_64: 12.8 MB
* arm64: 15.3 MB
* universal: 19.2 MB

@gxalpha or @PatTheMav if either of you would like to try to compile OBS and launch it against these to make sure the exclusion of dbus doesn't affect anything, let me know how it goes. Worst case, we can revert that change.

PR #105, opened 2022-05-19
FFmpeg: Remove Media Foundation dependency

flaeri:
### Description

This change instructs the ffmpeg build script to exclude Media Foundation as a dependency, thru `--disable-mediafoundation`

### Motivation and Context

A few versions back we bumped FFMPEG to a version that enables Media Foundation by default. I dont believe we need to rely on Media Foundation for the time being.

For Windows 10/11 "N" versions, which ships without Media Foundation installed, it will completely prevent the user from running OBS Studio.

A fairly large number of users run Windows N are coming by the support channels asking for help.
Manually installing Media Foundation can be quite confusing and difficult for a lot of users.

### How Has This Been Tested?

Built the modified FFMPEG thru our build scripts.
Verified that the avcodec-58.dll no longer depends on mfplat.dll.
Dropped the avcodec-58.dll into a release OBS, on Windows 10 N 22H1, and confirmed it can now launch, even without Media Foundation (mfplat.dll) installed.

Tested features I know uses ffmpeg:

* Media source, single h264 file
* Custom FFmpeg output, x264

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/windows_deps.yml
+++ b/.github/workflows/windows_deps.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         arch: ['x86', 'x86_64']
     env:
-      CACHE_REVISION: '01'
+      CACHE_REVISION: '02'
       MBEDTLS_VERSION: '2.24.0'
       MBEDTLS_HASH: '523f0554b6cdc7ace5d360885c3f5bbcc73ec0e8'
       PTHREAD-WIN32_VERSION: '2.10.0.0'
--- a/CI/windows/build_ffmpeg.sh
+++ b/CI/windows/build_ffmpeg.sh
@@ -75,6 +75,7 @@ _build_product() {
         --enable-libvpx \
         --enable-libsrt \
         --enable-librist \
+        --disable-mediafoundation \
         --$SWITCH-libaom \
         --$SWITCH-libsvtav1 \
         --disable-debug \


RytoEX:
As an aside, you'll have to increase the CACHE_REVISION to force CI to actually build these since no version numbers are changing. Otherwise, it will attempt to use a cache and skip building.

RytoEX:
> > As an aside, you'll have to increase the CACHE_REVISION to force CI to actually build these since no version numbers are changing. Otherwise, it will attempt to use a cache and skip building.
> 
> Forgive me @RytoEX, I am very much a novice when it comes to CI. I bump this one?
> 
> https://github.com/obsproject/obs-deps/blob/3cbc7b2e3c5973ec17c4f8b617a0207e555cad24/.github/workflows/main.yml#L25
> 
> Or this one? Or both?
> 
> https://github.com/obsproject/obs-deps/blob/214c5ce0abf51525d39f7072f468337a12723e10/.github/workflows/windows_deps.yml#L23

Right now, the main.yml file is for macOS deps only, and the windows_deps.yml file is for Windows deps. Only change the one for Windows.

RytoEX:
> For Windows 10/11 "N" versions, which ships without Media Foundation installed, it will completely prevent the user from running OBS Studio.

To me, this sounds like a bug should be filed upstream with FFmpeg, but I would need a test case to provide. To do that, we would need to know exactly where in OBS this is failing so we can write a failing test case that is minimal instead of just pointing at the OBS codebase.

I'll merge this once CI is done, if only to prevent OBS from being completely broken on Windows N installations. If we ever decide to utilize Media Foundation via FFmpeg though, this will break on Windows N editions again.

PR #108, opened 2022-05-30
Properly support multiple VS versions on Windows

WizardCM:
### Description

After failing to build obs-deps on Windows, coordinated with @PatTheMav to find a solution that'd fix building.

If multple copies of VIsual Studio are detected, sort the list and choose the oldest version installed.

### Motivation and Context

If more than one copy of Visual Studio is installed, `MSFT_VSInstance` will return an array of objects, causing the rest of the build script to fail.

Additionally, detours failed to build as it bypassed the Find-VisualStudio function, resulting it bypassing the included fix.

### How Has This Been Tested?

* Attempt to build obs-deps on a machine where more than one version/copy of Visual Studio is installed.
* Tested with 2019 and 2022 installed. With these changes, 2019 is selected.

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Invoke-DevShell.ps1
+++ b/utils.pwsh/Invoke-DevShell.ps1
@@ -29,9 +29,11 @@ function Invoke-DevShell {
         . $PSScriptRoot/Logger.ps1
     }
 
-    if ( ! ( Test-Path variable:VisualStudioData ) ) {
-        $VisualStudioData = Get-CimInstance MSFT_VSInstance
+    if ( ! ( Test-Path function:Find-VisualStudio ) ) {
+        . $PSScriptRoot/Setup-Target.ps1
     }
+
+    $VisualStudioData = Find-VisualStudio
 
     $DevShellCommand =
 @"
--- a/utils.pwsh/Setup-Target.ps1
+++ b/utils.pwsh/Setup-Target.ps1
@@ -78,6 +78,10 @@ function Find-VisualStudio {
 
     $VisualStudioData = Get-CimInstance MSFT_VSInstance
 
+    if ( $VisualStudioData.GetType() -eq [object[]] ) {
+        $VisualStudioData = ($VisualStudioData | Where-Object {$_.Version -ge 16} | Sort-Object -Property Version)[0]
+    }
+
     if ( ! ( $VisualStudioData ) -or ( $VisualStudioData.Version -lt 16 ) ) {
         $ErrorMessage = @(
             "A Visual Studio installation (2019 or newer) is required for this build script.",


WizardCM:
Pat has mentioned he's unsure whether this should prioritise the older or the newer Visual Studio. In an ideal world, developers should remove outdated/unused/legacy Visual Studio installations.

I think for a simple fix like this PR, we should prioritise newer versions.

Ideally the proper fix should be a little more complex. We should have a "recommended" version that we try and spot first (such as 16.x) and fall back to other versions.

utils.pwsh/Invoke-DevShell.ps1
```diff
@@ -29,10 +29,12 @@ function Invoke-DevShell {
         . $PSScriptRoot/Utils-Logger.ps1
```
PatTheMav, line 29: While we're here, you might want to change this to import "Logger.ps1" - that line contains the old filename. It just doesn't break as long as you don't import "Invoke-DevShell.ps1" standalone.

utils.pwsh/Setup-Target.ps1
```diff
@@ -78,6 +78,15 @@ function Find-VisualStudio {
 
     $VisualStudioData = Get-CimInstance MSFT_VSInstance
 
+    if ( $VisualStudioData.GetType() -Eq [object[]] ) {
+        $VisualStudioData = ($VisualStudioData | Sort-Object -Property Version)[0]
+    }
+
+    $VisualStudioId = "Visual Studio {0} {1}" -f @(
```
PatTheMav, line 85: You don't need those lines - just the 3 lines above flattening a possible array to a single object.

PatTheMav (changes requested):
See inline comments.

jpark37:
> If multple copies of VIsual Studio are detected, sort the list and choose the oldest version installed.

Sometimes having older VS installed is unavoidable, e.g. need plugins that aren't available on newer VS. I would prioritize a fixed version, then new to old.

RytoEX (changes requested):
> Sometimes having older VS installed is unavoidable, e.g. need plugins that aren't available on newer VS. I would prioritize a fixed version, then new to old.

I agree with this earlier feedback. As we currently design for VS2019, we should prefer that in the selection, then cycle through installed versions, new to old. That said, I expect to transition obs-deps and obs-studio to VS2022 by the end of June.

Are the objects returned actually guaranteed to be in a specific order? Or are we assuming that they are?

RytoEX (approved):
I've force pushed changes to ensure that only VS2019+ versions are selected into the object. This ensures that we don't end up in a scenario where a user has VS2017 and VS2019 installed and the sorted object array ends up with VS2017 in slot 0, which results in the next check (`-or ( $VisualStudioData.Version -lt 16`) failing in a confusing manner.

PatTheMav (approved):

PR #115, opened 2022-06-11
utils.pwsh: Add Check-GitUser

RytoEX:
### Description

Add a utility to check if a repo has a git user configured. If not, configure one so that git can commit patches.

### Motivation and Context

Helps us handle cases where we need to commit a patch to prevent a repo from being considered dirty, which causes some build tools to error out.

### How Has This Been Tested?

Tested locally on Windows 10 in PowerShell 7.2.4.

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Check-GitUser.ps1
+++ b/utils.pwsh/Check-GitUser.ps1
@@ -1,0 +1,35 @@
+function Check-GitUser {
+    <#
+        .SYNOPSIS
+            Ensures that a git user is configured.
+        .DESCRIPTION
+            Checks whether a git user has been configured in a local repository. If not, configure a generic one.
+        .EXAMPLE
+            Check-GitUser
+    #>
+
+    if ( ! ( Test-Path function:Log-Information ) ) {
+        . $PSScriptRoot/Logger.ps1
+    }
+    if ( ! ( Test-Path function:Invoke-External ) ) {
+        . $PSScriptRoot/Invoke-External.ps1
+    }
+
+    Log-Information "Check git config for user..."
+
+    $GitUserEmail = git config --get user.email
+    if ( [string]::IsNullOrEmpty($GitUserEmail) ) {
+        Log-Information "Set git user.email..."
+        Invoke-External git config user.email "commits@obsproject.com"
+    } else {
+        Log-Information "Git user.email already set: ${GitUserEmail}"
+    }
+
+    $GitUserName = git config --get user.name
+    if ( [string]::IsNullOrEmpty($GitUserName) ) {
+        Log-Information "Set git user.name..."
+        Invoke-External git config user.name "OBS Project"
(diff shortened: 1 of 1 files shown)


PatTheMav:
Wouldn't it make more sense to wrap this into a function that implements the entire functionality of applying a specified patch to the repo and automatically commit with a default commit email and username?

RytoEX:
> Wouldn't it make more sense to wrap this into a function that implements the entire functionality of applying a specified patch to the repo and automatically commit with a default commit email and username?

While this might be more comprehensive/encompassing, I think that this would be a more complex endeavor than this approach. For context, this is currently how I use this function in my WIP Qt 5 build script:
```
function Patch {
    Log-Information "Patch (${Target})"
    Set-Location $Path

    $Patches | ForEach-Object {
        $Params = $_
        Safe-Patch @Params
    }

    Set-Location qtbase
    Check-GitUser
    git add .
    git commit -m "Simple fix for QTBUG-74606"
}
```

The motivation for this is because I currently use Qt's `init-repository` perl script from their super repository (https://github.com/qt/qt5/), which is used to initialize the submodules. That script will fail if any of the submodules are "dirty" (i.e., there are uncommitted changes/patches). In the previous build scripts for obs-deps, I did this same process (check git user and set if needed) for cross-compiling FFmpeg for Windows:
https://github.com/obsproject/obs-deps/blob/6366c22b731c31112449d1d323daafaf8e306e3c/CI/include/build_support.sh#L105-L130
https://github.com/obsproject/obs-deps/blob/6366c22b731c31112449d1d323daafaf8e306e3c/CI/windows/build_ffmpeg.sh#L18
https://github.com/obsproject/obs-deps/blob/6366c22b731c31112449d1d323daafaf8e306e3c/CI/windows/build_ffmpeg.sh#L44-L45

As an alternative, I'm tempted to do away with this and simply perform a `git clean` on the qt5 checkout to remove any uncommitted changes, which will also allow `perl init-repository` to succeed. While this approach (with Check-GitUser) is what I have working locally and on my fork's CI, I'm open to either to unblock my work on building Qt for Windows on CI so that we can move forward.

RytoEX:
> Fine be me, as long as it's idempotent (meaning you can run it again and again and again and it doesn't have negative impacts).

I retested my WIP qt5.ps1 script, and it does appear to be able to re-run any number of times with no repercussions because Setup calls `perl init-repository --module-subset="qtbase,qtimageformats,qtmultimedia,qtsvg,qtwinextras" --force` which resets the qtbase submodule to whatever the qt5 super-repo checkout is has it set to.

Discussed further off-thread and this seems fine to merge as-is since it works as intended and has no negative repercussions. Will make sure my local work is synced up here and merge after that.

PatTheMav (approved):

PR #121, opened 2022-07-05
windows-deps: Add patched swig binary with support for Py3 stable ABI

PatTheMav:
### Description
Adds swig to precompiled binaries for Windows with applied patch for Py3 stable ABI patch.

### Motivation and Context
Support for the Python 3 stable ABI allows Python scripting support in OBS to use a wider range of available Python 3 versions vs. being pinned to the version the scripting module is compiled with.

### How Has This Been Tested?
* Built and packaged Swig locally
* Compiled obspython with Swig compiled with these changes

### Types of changes

### Checklist:

### DIFF
--- a/Build-Dependencies.ps1
+++ b/Build-Dependencies.ps1
@@ -114,6 +114,26 @@ function Package-Dependencies {
     Push-Location -Stack BuildTemp
     Set-Location $ConfigData.OutputPath
 
+    Log-Information "Cleanup unnecessary files"
+    $Items = @(
+        "./bin/bison.exe"
+        "./bin/libiconv2.dll"
+        "./bin/libintl3.dll"
+        "./bin/m4.exe"
+        "./bin/pcre2*"
+        "./bin/regex2.dll"
+        "./cmake/pcre2*"
+        "./include/pcre2*"
+        "./lib/pcre2*"
+        "./lib/pkgconfig/libpcre2*"
+        "./man1/pcre2*"
+        "./man3/pcre2*"
+        "./share/bison"
+        "./share/doc/pcre2"
+    )
+
+    Remove-Item -ErrorAction 'SilentlyContinue' -Path $Items -Force -Recurse
+
     Log-Information "Package dependencies"
 
     $Params = @{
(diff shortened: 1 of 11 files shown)


jp9000:
Big hype. Can't wait until this is merged.

RytoEX:
Is this until SWIG merges the needed stable ABI PR, or does this also require an additional patch that it yet to be upstreamed (PR'd/merged)?

PatTheMav:
> Is this until SWIG merges the needed stable ABI PR, or does this also require an additional patch that it yet to be upstreamed (PR'd/merged)?

This just contains the PR re-based on master. It was easier to apply it as a single patch on master vs checking out the PR, then patching it further.

The patch is identical to the one on macOS apart from Windows line endings.

RytoEX:
There are two patches here though. Is the second one part of the aforementioned SWIG PR, or is it something completely separate that we have to maintain?

PatTheMav:
> There are two patches here though. Is the second one part of the aforementioned SWIG PR, or is it something completely separate that we have to maintain?

The second patch mainly fixes their CMakeLists script to work with the PCRE2 library as produced by PCRE itself. This is mainly due to Swig not maintaining a build script that works with the Windows libraries produced by PCRE out of the box.

Their docs otherwise require to pass the necessary variables manually (i.e. `-DCMAKE_C_FLAGS="/DPCRE2_STATIC" ^
-DCMAKE_CXX_FLAGS="/DPCRE2_STATIC" -DPCRE2_INCLUDE_DIR=%PCRE_ROOT%/include -DPCRE2_LIBRARY=%PCRE_ROOT%/lib/pcre2-8-static.lib`) which I found too brittle.

A fix would be to update Swig's CMakeLists file upstream to work with PCRE as produced by their build system out of the box on Windows.

RytoEX (approved):
PR looks fine to me. My one ask would be to upstream the additional patch to SWIG and hope that they merge it in the future.

RytoEX:
Per off-thread conversation with @PatTheMav , he will clean up and upstream the CMake patch at a later date. Merging.

PR #131, opened 2022-07-29
deps.windows: Update VulkanSDK to 1.3.216.0

RytoEX:
### Description

Continue using our own hosted Vulkan components.

The Vulkan SDK installer no longer includes the 32-bit components in it.
They are now an optional component downloaded during installation. Use
our own hosted copy of the components instead.

Additionally, this is effectively a double revert.

Revert "Revert "deps.windows: Update VulkanSDK to 1.3.216.0""

This reverts commit d876ba352ebc10f1af0c2eb1a22a6007975f0219.

### Motivation and Context

@jpark37 said we actually might need to update to avoid issues.

### Types of changes

### Checklist:


shijunz:
I have tow questions:
1 how about the vulkan for windows on ARM64? if support
2  why it is only the vulkan and vlc download, other module all use powershell to setup/build/install.

thanks.

RytoEX:
> I have tow questions: 1 how about the vulkan for windows on ARM64? if support 2 why it is only the vulkan and vlc download, other module all use powershell to setup/build/install.
> 
> thanks.

WoA is on our roadmap, but is not addressed by this PR. We don't even know if Game Capture will work on WoA, which is what we use Vulkan for. We only update Vulkan when a specific need arises because we don't want to break Game Capture.

This is still a PowerShell script. It is downloading the Vulkan SDK. We don't need to build anything. VLC doesn't need to build anything either, as we only require its headers.

Issue #135, opened 2022-09-22
GitHub action "Build macOS Qt6" fails on not finding GSS library


glikely:
### Operating System Info

macOS 12

### OBS Studio Version

Other

### OBS Studio Version (Other)

2022-08-02

### OBS Studio Log URL

https://github.com/obsproject/obs-deps/actions/runs/3099336070/jobs/5018349974

### Expected Behavior

Expected qt6 build to link without errors

### Current Behavior

The qt6 build fails with the following error:

```
[2969/3649] Linking CXX shared library lib/QtNetwork.framework/Versions/A/QtNetwork
FAILED: lib/QtNetwork.framework/Versions/A/QtNetwork 
: && /Applications/Xcode_13.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -w -pipe -stdlib=libc++ -DNDEBUG -O2 -arch x86_64 -arch arm64 -isysroot /Applications/Xcode_13.4.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk -mmacosx-version-min=10.14 -dynamiclib -Wl,-headerpad_max_install_names -Wl,-undefined,error -fapplication-extension -compatibility_version 6.0.0 -current_version 6.3.1 -o lib/QtNetwork.framework/Versions/A/QtNetwork -install_name @rpath/QtNetwork.framework/Versions/A/QtNetwork src/network/CMakeFiles/Network.dir/Network_autogen/mocs_compilation.cpp.o 
[...]
[...]

PR #136, opened 2022-09-22
deps.qt: Fix macOS build by disabling VCPKG

glikely:
The macOS Qt6 Universal build fails on GitHub actions due to a suspected bug in vcpkg. Disable using vcpkg until a proper fix is found.

This is an RFC pull request. I don't know if it is the best solution and I'm looking for feedback.

Fixes: #135

### How Has This Been Tested?
Has had some minimal testing on an unattached repo (https://github.com/glikely/obs-deps-2. Will get a full test when the CI Action is approved for this pull request.

### Types of changes

### Checklist:

### DIFF
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -91,6 +91,7 @@ config() {
   if (( ${+commands[ccache]} )) common_cmake_flags+=(-DQT_USE_CCACHE=ON)
 
   if [[ ${CPUTYPE} != "${arch}" && ${host_os} == 'macos' ]] {
+    unset VCPKG_ROOT
     if ! /usr/bin/pgrep -q oahd; then
       local -A other_arch=(arm64 x86_64 x86_64 arm64)
       common_cmake_flags+=(-DCMAKE_OSX_ARCHITECTURES="${CPUTYPE};${other_arch[${CPUTYPE}]}")


PatTheMav:
Judging by this:

https://github.com/qt/qtbase/blob/3cb00e415d72f466a1f235490cfae029169ce7ae/cmake/QtAutoDetect.cmake#L163

Unsetting the `VCPKG_ROOT` environment variable might be the only viable fix, as Qt runs this automatic discovery.

RytoEX:
You don't need the prefix "workaround: ". Please review our commit history for previously used commit prefixes.

glikely:
> You don't need the prefix "workaround: ". Please review our commit history for previously used commit prefixes.

Done

RytoEX:
I guess my only misgiving here is if VCPKG_ROOT is set locally (I don't know how likely this is), then we're unconditionally removing it and not restoring the previously set value. We should try to be less destructive to local build environments. @PatTheMav thoughts?

glikely:
Unsettling VCPKG_ROOT only has effect within running the scope of the script. The local env should not be affected.

RytoEX:
> Unsettling VCPKG_ROOT only has effect within running the script. The local env should not be affected.

Ah, in that case then, it's fine. Will leave to @PatTheMav to approve/merge since it's macOS.

PR #137, opened 2022-09-23
deps.qt: Add QtSerialPort

glikely:
### Description
At least one plugin (obs-ptz) requires the QtSerialPort library. Add it to the base set of OBS Qt dependencies so that the plugin doesn't need to supply its own copy which simplifies the build and also ensures the library versions remain in-sync with each other.

### How Has This Been Tested?
Tested with the obs-ptz plugin. Builds successfully on Windows and macOS

### Types of changes

### Checklist:


RytoEX:
As we do not use QtSerialPort in OBS Studio itself, and obs-deps is for providing dependencies that OBS Studio uses, we'll be closing this PR. Thank you for your understanding.

PR #138, opened 2022-10-15
Update GitHub Actions for set-output deprecation

RytoEX:
### Description

GitHub Actions has deprecated set-output. Replace usages of set-output in stdout with the new syntax to save the output to the new environment variable.

See:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### Motivation and Context

Don't want deprecation warnings hiding real issues.

### How Has This Been Tested?

It hasn't. This was written by following documentation. A CI run will be the test.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/build-ffmpeg/action.yml
+++ b/.github/actions/build-ffmpeg/action.yml
@@ -37,7 +37,7 @@ runs:
         esac
 
         ffmpeg_dep_hash=$(cat ${{ github.workspace }}/deps.ffmpeg/*.zsh | sha256sum | cut -d " " -f 1)
-        echo "::set-output name=hash::${ffmpeg_dep_hash}"
+        echo "hash=${ffmpeg_dep_hash}" >> $GITHUB_OUTPUT
 
     - name: Restore FFmpeg Dependencies from Cache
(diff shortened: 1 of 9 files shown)


PatTheMav (approved):
Just checked, and according to GitHub's docs something like `"{name}=value" >> $env:GITHUB_OUTPUT` is the right pattern (but only in PowerShell 5.1 and above), but the existing changes should be fine.

RytoEX:
> Just checked, and according to GitHub's docs something like `"{name}=value" >> $env:GITHUB_OUTPUT` is the right pattern (but only in PowerShell 5.1 and above), but the existing changes should be fine.

Could you point me to the specific documentation example?

PatTheMav:
> > Just checked, and according to GitHub's docs something like `"{name}=value" >> $env:GITHUB_OUTPUT` is the right pattern (but only in PowerShell 5.1 and above), but the existing changes should be fine.
> 
> Could you point me to the specific documentation example?

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions?tool=powershell#setting-an-output-parameter

PR #139, opened 2022-10-19
deps.windows: Add Jansson to Windows deps

RytoEX:
### Description

Add Jansson 2.14 to the Windows deps. Mimic our existing in-tree build options as much as possible, except don't override JANSSON_STATIC_CRT. The default value is OFF, which in turn means /MT and /MTd are NOT set, which is seemingly what we want. Our in-tree build overrides JANSSON_STATIC_CRT to ON, but then also forces /MT and /MTd to NOT be set. JANSSON_STATIC_CRT only controls whether or not those flags are set, so just leave it OFF.

### Motivation and Context

Jansson adds 20-30 seconds to the CMake configuration time of my local builds, and adds 40-60 seconds on CI builds. If we prebuild it, we can avoid that needless expenditure of time. Want to pull in-tree deps out of the obs-studio tree where feasible.

### How Has This Been Tested?

Tested locally and I can build Release obs-studio just fine. However, trying to build in Debug results in an LNK4098 error:
```
15>   Creating library C:/code/obsproject/obs-studio/build64/libobs/Debug/obs.lib and object C:/code/obsproject/obs-studio/build64/libobs/Debug/obs.exp
15>LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
15>LINK : error LNK1218: warning treated as error; no output file generated
15>Done building project "libobs.vcxproj" -- FAILED.

21>   Creating library C:/code/obsproject/obs-studio/build64/plugins/rtmp-services/Debug/rtmp-services.lib and object C:/code/obsproject/obs-studio/build64/plugins/rtmp-services/Debug/rtmp-services.exp
21>LINK : warning LNK4098: defaultlib 'MSVCRT' conflicts with use of other libs; use /NODEFAULTLIB:library
21>LINK : error LNK1218: warning treated as error; no output file generated
21>Done building project "rtmp-services.vcxproj" -- FAILED.
```
We have previously suppressed this warning in https://github.com/obsproject/obs-studio/pull/6774 and https://github.com/obsproject/obs-studio/pull/6723. I don't know if that's what we should do in this case as well, or if there's a more correct solution.
### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/30-jansson.ps1
+++ b/deps.windows/30-jansson.ps1
@@ -1,0 +1,66 @@
+param(
+    [string] $Name = 'jansson',
+    [string] $Version = '2.14',
+    [string] $Uri = 'https://github.com/akheron/jansson.git',
+    [string] $Hash = '684e18c927e89615c2d501737e90018f4930d6c5'
+)
+
+function Setup {
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
+}
+
+function Clean {
+    Set-Location $Path
+    if ( Test-Path "build_${Target}" ) {
+        Log-Information "Clean build directory (${Target})"
+        Remove-Item -Path "build_${Target}" -Recurse -Force
+    }
+}
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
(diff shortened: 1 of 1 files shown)


PatTheMav:
From what I could tell this switch needs to match the one used by the host executable/library, and libobs uses `/MDd`, but Jansson's CMake only explicitly sets `/MTd` or nothing at all.

PR #140, opened 2022-10-29
Add support for Windows ARM64

tommyvct:
### Description
Add support for Windows ARM64

### Motivation and Context
Make OBS Studio great on Windows ARM64 devices.

### How Has This Been Tested?
Still need further integrated test. 

### Types of changes

### Checklist:


.github/actions/build-windows-qt/action.yml
```diff
@@ -51,6 +51,17 @@ runs:
       shell: pwsh
       run: ./Build-Dependencies.ps1 -Dependencies "Qt${{ inputs.qtVersion }}" -SkipBuild -SkipUnpack -Target ${{ inputs.target }} -Configuration ${{ inputs.config }} -Shared
 
+    - name: Install Qt
```
PatTheMav, line 54: Don't use third party actions, if you need a host-version of Qt, wait for the associated x64 job to finish, then use that artifact.

deps.ffmpeg/20-opus.zsh
```diff
@@ -57,6 +57,20 @@ config() {
         -DOPUS_FORTIFY_SOURCE=OFF
       )
       ;;
+    windows-arm64)
+      args+=(
+        -DOPUS_STACK_PROTECTOR=OFF
+        -DOPUS_FORTIFY_SOURCE=OFF
+        -DRUNTIME_CPU_CAPABILITY_DETECTION=OFF
+        -DOPUS_DISABLE_INTRINSICS=ON
+        # TODO: llvm-mingw ARM64 seh __try
```
PatTheMav, line 66: Remove unused code for PRs instead of using comments.

deps.ffmpeg/30-libvpx.zsh
```diff
@@ -81,9 +81,14 @@ config() {
         args+=(--target="${arch}-darwin20-gcc")
       }
       ;;
-    windows-x*)
+    windows-*)
       cross_prefix="${target_config[cross_prefix]}-w64-mingw32-"
-      args+=(--enable-runtime-cpu-detect --target="${target_config[gcc_target]}")
+      args+=(--target="${target_config[gcc_target]}")
```
PatTheMav, line 86: Use separate `windows-arm64` branch, there is no need to check the target again when you could've matched the target in the `case` statement.

deps.ffmpeg/60-srt.zsh
```diff
@@ -61,16 +61,21 @@ config() {
...
       args+=(
         -DUSE_OPENSSL_PC=OFF
         -DCMAKE_CXX_FLAGS="-static-libgcc -static-libstdc++ -w -pipe -fno-semantic-interposition"
         -DCMAKE_C_FLAGS="-static-libgcc -w -pipe -fno-semantic-interposition"
-        -DCMAKE_SHARED_LINKER_FLAGS="-static-libgcc -static-libstdc++ -L${target_config[output_dir]}/lib -Wl,--exclude-libs,ALL"
         -DSSL_LIBRARY_DIRS="${target_config[output_dir]}/lib"
         -DSSL_INCLUDE_DIRS="${target_config[output_dir]}/include"
       )
 
+      if [[ ${target} == "windows-x"* ]]; then
```
PatTheMav, line 73: See above, use distinct `case` match for arm64 to cleanly separate architectures.

deps.ffmpeg/70-nv-codec.zsh
```diff
@@ -11,11 +11,21 @@ local targets=('windows-x*')
 
 ## Build Steps
 setup() {
+  if [[ ${target} == "windows-arm64" ]]; then
```
PatTheMav, line 14: This should not be necessary as the `targets` variable above explicitly limits the dependency to `windows-x*` targets, which should not match `windows-arm64`.

deps.ffmpeg/80-amf.zsh
```diff
@@ -11,12 +11,22 @@ local targets=('windows-x*')
 
 ## Build Steps
 setup() {
+  if [[ ${target} == "windows-arm64" ]]; then
```
PatTheMav, line 14: See comments on `70-nv-codec.zsh`.

.github/actions/build-windows-qt/action.yml
tommyvct: okay.

deps.ffmpeg/99-ffmpeg.zsh
```diff
@@ -158,7 +158,7 @@ config() {
       )
 
       if (( ! shared_libs )) {
-        ff_ldflags+=(-Wl,-Bstatic -pthread)
+        ff_ldflags+=(-Wl,-Bstatic -pthread -Wl,-Bdynamic)
```
PatTheMav, line 161: If `shared_libs` is explicitly disabled, do not build dynamic libraries. This setting goes against the explicit user choice.

deps.ffmpeg/20-opus.zsh
tommyvct: ok

deps.qt/qt5.ps1
```diff
@@ -25,6 +25,11 @@ param(
 # path should also be kept short. This avoids issues with too long file paths in the compilation phase.
 
 function Setup {
+    if (${Target} -eq "arm64") {
```
PatTheMav, line 28: This should be implemented the same way as in the `zsh`-framework, with a local variable setting an allow-list and the dispatcher checking the the current target architecture is in that list.

Applies to all the similar changes below.

deps.qt/qt6.ps1
```diff
@@ -82,14 +86,19 @@ function Configure {
         $QtBuildConfiguration = '-release'
     }
 
-    $BuildCommand = "..\..\..\qt6\configure -opensource -confirm-license ${QtBuildConfiguration} -nomake examples -nomake tests -schannel -no-dbus -no-freetype -no-icu -no-openssl -no-feature-androiddeployqt -no-feature-pdf -no-feature-printsupport -no-feature-qmake -no-feature-sql -no-feature-testlib -no-feature-windeployqt -DQT_NO_PDF -prefix ${BuildPath}"
+    $BuildCommand = "..\..\..\qt6\configure "
```
PatTheMav, line 89: Don't directly append to strings, use an array with the parameters as tokens and append to the array, finally joining the array with a glue or via splatting.

deps.qt/qt6.ps1
```diff
@@ -82,14 +86,19 @@ function Configure {
...
+    }
+    
+    $BuildCommand += "-opensource -confirm-license ${QtBuildConfiguration} -nomake examples -nomake tests -schannel -no-dbus -no-freetype -no-icu -no-openssl -no-feature-androiddeployqt -no-feature-pdf -no-feature-printsupport -no-feature-qmake -no-feature-sql -no-feature-testlib -no-feature-windeployqt -DQT_NO_PDF -prefix ${BuildPath}"
 
     $Params = @{
         BasePath = (Get-Location | Convert-Path)
         BuildPath = "."
         BuildCommand = "${BuildCommand}"
         Target = $Target
-        HostArchitecture = $Target
```
PatTheMav, line 92: This is necessary to build x86 artifacts on x64 hosts, do not remove.

deps.windows/60-python-arm64.ps1
```diff
@@ -0,0 +1,65 @@
+param(
```
PatTheMav, line 1: Use PyEnv for this - arm64 support is in an alpha state (iirc) and PyEnv support is forthcoming. Until this version of Python is stable enough to ship with OBS, PyEnv support should be available.

Until then, skip building Python for obs-deps.

deps.ffmpeg/99-ffmpeg.zsh
tommyvct: No, it should not be removed. 

https://github.com/mstorsjo/llvm-mingw/issues/311#issuecomment-1312589456

This gonna be a long read, but the TLDR is the default state of the linker args for external libs without `-Wl,-Bstatic` is dynamic. By not switching it back to dynamic causes problems, at least on llvm toolchain. Regardless of the toolchain used, this should be rigorous.

utils.zsh/20-host.zsh
```diff
@@ -56,7 +56,7 @@ _setup_macos() {
 }
 
 _setup_linux() {
-  autoload -Uz is-at-least check_deps check_git check_nasm check_cross
+  autoload -Uz is-at-least check_deps check_git check_nasm check_cross check_cross_windows_arm64
```
PatTheMav, line 59: Rename `check_cross` to `check_cross_windows_x86` to make it clear it's separated by architecture.

utils.zsh/functions/check_cross_windows_arm64
```diff
@@ -0,0 +1,57 @@
...
+# if (( ! ${+commands[apt-get]} )) {
+#   log_error 'No apt-get command found. Please install apt'
+#   return 2
+# } else {
+#   log_debug "Apt-get located at ${commands[apt-get]}"
+# }
+
+# local -a _quiet=('' '--quiet')
+
+# if [[ ! -f "`pwd`/llvm-mingw-20220906-ucrt-ubuntu-18.04-x86_64/bin/aarch64-w64-mingw32-clang" &&
```
PatTheMav, line 13: Do not use backticks for command invocations, use the proper `$()` pattern.

utils.zsh/functions/check_cross_windows_arm64
```diff
@@ -0,0 +1,57 @@
+autoload -Uz log_info log_status log_error log_debug log_warning
+
+# log_debug 'Checking for apt-get...'
```
PatTheMav, line 3: Remove code if it's not needed anymore.

utils.zsh/functions/check_cross_windows_arm64
```diff
@@ -0,0 +1,57 @@
...
+    fi
+    log_warning 'Add llvm-mingw toolchain to PATH'
+    export PATH="`pwd`/llvm-mingw-20220906-ucrt-ubuntu-18.04-x86_64/bin:$PATH"
+  ;;
+  macos)
+    if [[ ! -f "`pwd`/llvm-mingw-20220906-ucrt-macos-universal/bin/aarch64-w64-mingw32-clang" ]]; then
+      log_warning 'Downloading llvm-mingw toolchain'
+      wget https://github.com/mstorsjo/llvm-mingw/releases/download/20220906/llvm-mingw-20220906-ucrt-macos-universal.tar.xz
+      log_warning 'Expanding llvm-mingw toolchain'
+      tar -xf llvm-mingw-20220906-ucrt-macos-universal
```
PatTheMav, line 52: Do not extract into project directory - if additional 3rd party tools are installed, they should be put in the correct places on the system, in this case either `/usr/bin/` or `/usr/local/bin`, which also would not make `PATH` adjustments necessary.

.github/actions/build-windows-qt/action.yml
PatTheMav: Let us know if that is not feasible as-is - in doubt we can and should amend the current Qt build workflow to generate what this workflow needs. Also I don't think I have seen associated steps in the build scripts ensuring that a host installation of Qt is available to cross compile it?

deps.ffmpeg/99-ffmpeg.zsh
PatTheMav: As that is only necessary for the `llvm`-based toolchain, it should be limited to that and not removed for other architectures.

utils.zsh/functions/check_cross_windows_arm64
tommyvct: According to https://github.com/mstorsjo/llvm-mingw/issues/250, it should be used in this way.

This toolchain is unpacked, if we just extract the toolchain to anywhere that requires `sudo`, that would be a mess. Keep in mind that CI is not the only end user of this script. You don't want to ruin other people's machine.

utils.zsh/20-host.zsh
tommyvct: done

deps.qt/qt6.ps1
tommyvct: disabled

deps.windows/60-python-arm64.ps1
tommyvct: disabled

deps.ffmpeg/30-libvpx.zsh
tommyvct: done

deps.ffmpeg/60-srt.zsh
tommyvct: That's a lot of duplicate codes

deps.ffmpeg/80-amf.zsh
tommyvct: done

.github/actions/build-windows-qt/action.yml
tommyvct: I think that's possible in theory. I will need to dig a little bit about the CI though.

tommyvct:
I'll leave the rest later.

deps.ffmpeg/60-srt.zsh
```diff
@@ -61,16 +61,28 @@ config() {
...
       args+=(
         -DUSE_OPENSSL_PC=OFF
         -DCMAKE_CXX_FLAGS="-static-libgcc -static-libstdc++ -w -pipe -fno-semantic-interposition"
         -DCMAKE_C_FLAGS="-static-libgcc -w -pipe -fno-semantic-interposition"
-        -DCMAKE_SHARED_LINKER_FLAGS="-static-libgcc -static-libstdc++ -L${target_config[output_dir]}/lib -Wl,--exclude-libs,ALL"
         -DSSL_LIBRARY_DIRS="${target_config[output_dir]}/lib"
         -DSSL_INCLUDE_DIRS="${target_config[output_dir]}/include"
       )
 
+      args+=(-DCMAKE_SHARED_LINKER_FLAGS="-static-libgcc -static-libstdc++ -L${target_config[output_dir]}/lib -Wl,--exclude-libs,ALL")
```
PatTheMav, line 73: Unnecessary code - don't use a second call to append to the array, just revert the change.

deps.ffmpeg/60-srt.zsh
```diff
@@ -61,16 +61,28 @@ config() {
...
+    windows-arm64)
+      args+=(
+        -DUSE_OPENSSL_PC=OFF
+        -DCMAKE_CXX_FLAGS="-static-libgcc -static-libstdc++ -w -pipe -fno-semantic-interposition"
+        -DCMAKE_C_FLAGS="-static-libgcc -w -pipe -fno-semantic-interposition"
+        -DSSL_LIBRARY_DIRS="${target_config[output_dir]}/lib"
+        -DSSL_INCLUDE_DIRS="${target_config[output_dir]}/include"
+      )
+        
+      args+=(-DCMAKE_SHARED_LINKER_FLAGS="-static-libgcc -static-libstdc++ -L${target_config[output_dir]}/lib")
```
PatTheMav, line 85: Same as above.

deps.ffmpeg/60-srt.zsh
```diff
@@ -140,6 +152,11 @@ fixup() {
       }
 
       autoload -Uz restore_dlls && restore_dlls
+      
+      # TODO: To be removed once upgraded to 1.6.0
+      # ref: https://github.com/Haivision/srt/pull/2536
+      log_info "Fixup pkgconfig"
+      sed -i 's/-l-l:libunwind.a //g' ${target_config[output_dir]}/lib/pkgconfig/srt.pc
```
PatTheMav, line 159: Use a patch file with proper link to the GitHub issue and corresponding file name instead.

deps.qt/qt5.ps1
```diff
@@ -103,13 +118,18 @@ function Configure {
         BuildPath = "."
         BuildCommand = "${BuildCommand}"
         Target = $Target
-        HostArchitecture = $Target
+        HostArchitecture = Get-HostArchitecture
```
PatTheMav, line 121: @RytoEX is this correct? It seems wrong to me because that would force the DevShell to use the host architecture compiler (e.g. x64) to build another target (i.e. x86) and would fail.

deps.qt/qt6.ps1
```diff
@@ -34,6 +34,10 @@ function Setup {
         '--force'
     )
     Invoke-External perl init-repository @Options
+
+    if ( (${Target} -eq "arm64") -and ($Env:Qt6_DIR -eq $null) ) {
```
PatTheMav, line 38: Do not rely on an environment variable, add an explicit build script parameter that needs to be supplied upon invocation (e.g. `QtHostInstall`) which is checked for and used instead.

utils.pwsh/Get-HostArchitecture.ps1
```diff
@@ -0,0 +1,17 @@
...
+    #>
+
+    switch ((Get-CimInstance -ClassName CIM_OperatingSystem).OSArchitecture) {
+        "ARM 64-bit Processor" { return "ARM64" }
+        "64-bit" { return "x64" }
+        "32-bit" { return "x86" }
+    }
+
+    return "Unknown"    
+}
```
PatTheMav, line 17: Missing newline at end of file.

utils.zsh/functions/check_cross_windows_arm64
PatTheMav: We already install mingw64 and other build-time dependencies in the system per Linux standards, llvm-mingw should be installed in the same way. That would also remove the requirement to add an additional directory to the `gitignore` file (which is already the first sign that it's not a good practise).

We also don't support cross compiling on macOS (runners are too slow and not available enough for that to make sense), so you can remove that entirely.

More things:
* Use `curl` (which is the canonical tool we use to fetch things and available out of the box on macOS) instead of `wget`.
* Use the `extract` function from the zsh-framework to extract the archive

If you want/need temporary directories, use the `mktemp` function.

deps.qt/qt5.ps1
RytoEX: HostArchitecture appears to set the host toolchain rather than specifying the what the host OS architecture. While the dev shell may be able to cross-compile, and will be necessary for arm64 on CI until there are WoA runners, it's best to avoid cross-compiling when possible, and we don't want to force x86 cross-compiles from the x64 host toolchain. In my experience, some projects also don't handle such instances well. They may not build, or they may build, but incorrectly.

tommyvct:
This is painful.

tommyvct:
I have absolutely no idea why the CI failed. Everything compiles just fine on my end.

pbo-linaro:
Hi everyone,

at Linaro, we are actively working to help some projects to [support Windows on Arm](https://www.linaro.org/windows-on-arm/) and we would like to help OBS Studio to get ported on this platform.

It seems like a lot of work has been already done (thanks to @tommyvct - nice nickname by the way).
Is there a specific part where we can help? (automate CI for build dependencies, refactor an existing PR, anything else)

Thanks,
Pierrick

derrod:
I believe @PatTheMav has also made some progress with getting OBS to run on Windows on ARM (in a VM on Apple Silicon at least) as part of the build system rework that's currently being finalised, so he'll probably have some more insight on the current status of things. This PR is a bit outdated, and will likely be superseeded.

One major concern I am aware of is the lack of hardware encoding support as this only seems to be available via MediaFoundation. And it would probably be quite important to have given the performance - or rather lack thereof - of pretty much all Windows on ARM machines.

tommyvct:
> Is there a specific part where we can help? (automate CI for build dependencies, refactor an existing PR, anything else)


First, thanks for your intention to help, @pbo-linaro !

I was pretty busy working on my degrees in the past 6 months, and after a quick smoke test that I did about a month ago, my previous work fell flat with the current OBS Studio codebase. One of the notable changes that caused the this is the CMake overhaul glued NVENC to the libobs and requires some effort to separate, which the ARM64 build need to get rid of them.

Another major roadblocker, as @derrod said, is the lack of hardware encoding. OBS Studio used to have a Media Foundation encoding backend, but it was removed long ago in favor to vendor-specific solutions like AMF, QSV, and NVENC. The encoding and streaming performance with the version of OBS I compiled before, using software encoding, on a Surface Pro X SQ1, is perfectly usable, even for some light gaming. 

Qualcomm chips like 8/7cx series are based on Snapdragon chips for smartphones, which is known to have a hardware encoder. However, after extensive research, I cannot find any usable information about how to utilize the hardware. I did found something remotely close to somewhat useful: there is a PDF document claiming that the encoder supports Windows 8 and some example, indicating the documentation was obsolete. What's worse is the SDK is not easily accessible, at least without contacting and paying a lot of money. I know Linaro is not Qualcomm, and it's gonna be most probably impossible, but can you find us some recent and usable documentation on the Qualcomm hardware encoder? Thanks!

tommyvct:
And for the CI/Automation part, @PatTheMav has full control of this part. I'm simply extending and replicating his work to make OBS Studio work on Windows on ARM.

pbo-linaro:
Thanks for your feedback @tommyvct and @derrod.

Good news to hear it's still a work in progress, and that @PatTheMav is on track to enable it.

Concerning the hardware encoding support, I'll contact engineers we work with in Qualcomm, and see what we can do about it.

PatTheMav:
If we get around to implement ARM64 on Windows support (which will have to be dependent on whether contributors and volunteer supporters feel comfortable with the added support burden given the still somewhat limited availability of Windows for ARM64), we'd probably base it off the native Windows compilation toolchain for FFmpeg instead of cross-compilation.

In my tests there are the following remaining issues:

* libvpx fails to compile with MSVC's native toolchain:
    * `libvpx\vpx_dsp\arm\fdct4x4_neon.c(55,52): error C2078`
* x264 fails to produce ARM64 assembly by default. The `gas-preprocessor` perl script used by it are 9 years old and calls `cpp` to do actual preprocessing (even if `cl.exe` is used). FFmpeg's fork of the preprocessor fixes this, so the x264's version needs to be replaced with it to build on Windows 
* FFmpeg itself just requires the same `gas-preprocessor` script to be downloaded and made available in the PATH for the compilation

On top of that I think that MediaFoundation support needs to be enabled again, as I would expect Windows on ARM to be running on SOC-based systems (either Qualcomm-based SOCs, Microsoft's own, or Apple Silicon), so much like `VideoToolbox` it will be the official way to get hardware acceleration working on Windows.

And as correctly pointed out this will also require the obs-studio Windows build system to require MediaFoundation on ARM64 again and ignore AMF/NVEnc and other x64-specific features.

PatTheMav:
> @PatTheMav In previous messages, it was suggested your were working on the rewrite of the build system. Is that the case? If yes, is that close to completion?
> 
> I'm trying to understand when it will be possible to contribute to a "stable" state, to avoid redundant/lost work.

That'd be https://github.com/obsproject/obs-deps/pull/186.

tommyvct:
There's an update from the Qt side: https://www.qt.io/blog/qt-for-windows-on-arm

TL:DR is they have yet to decide whether ARM64 or ARM64EC will be their "native ARM64 architecture". 


My previous work was all around the ARM64 ABI. If Qt changed their preferred ABI to ARM64EC, then most of my work (not my research and experience, though) would be in vain. 

I did some experiments with the dependencies with the ARM64EC ABI before, and it seems that dependencies that's using anything other than msbuild or CMake, or heavily utilize intrinsics, will have major problems. The ARM64EC toolchain self-identifies the target machine as `_M_X64`, and the compiler and the linker don't have the same invocation behaviour as the x64 one. All in all, the program we are compiling will have some serious identity crisis, unless they are particularly patched for ARM64EC. 

ARM64EC doesn't necessarily mean all trouble and no benefit, though. Say if a plugin like LuaJIT still doesn't have Windows ARM64 support, ARM64EC version of OBS Studio could just use the x64 version of DLL just fine. 

In the end of the day, the choice between ARM64 or ARM64EC depends on 1) the usage and expectation from the WoA users and 2) the thoughts from the plugin developers and 3) the Qt company. 

I personally would lean towards ARM64, since that what we have on hand that is working(?) for now.

PatTheMav:
From our POV the requirements are quite simple:

* It has to compile when we throw it at MSVC's arm64 compiler

That's all there is to it. We will not cross-compile or use code generated by MINGW toolchains, it has to be compiled with MSVC. This works for almost all dependencies (as I outlined above), `libvpx` being the only one I wasn't able to figure out a fix for yet.

tommyvct:
> * It has to compile when we throw it at MSVC's arm64 compiler

The question I'm talking about is **which** compiler or toolset we should use, ARM64 or ARM64EC. I would say just stay with ARM64, since we don't have nasty legacies like some old commercial softwares.

driver1998:
If OBS supports va-api on Windows, we might be able to get it working on Windows with the new vaOn12 driver (which wraps va-api to DirectX 12). Not sure about how it works on Qualcomm platform though. (DirectX 12 in general is supported here, but not sure about DirectX 12 video).

driver1998:
> The question I'm talking about is which compiler or toolset we should use, ARM64 or ARM64EC.

I would say ARM64. We are mostly dealing with open source plugins right now, so the benefits of ARM64EC is not very appealing. (We can always port those plugins over, it's not like we haven't dealt with Linux ARM64 or anything).
 
Also with ARM64 we can use alternative toolchains like MinGW, even though we don't use it for official releases, some plugins might want or require it.

tommyvct:
It looks that vaon12 is more promising than the media foundation encoder.

driver1998:
Looks like vaOn12 is decode only on 8cx Gen 2:

```
$ vainfo
Trying display: win32
libva info: VA-API version 1.18.0
libva info: User environment variable requested driver 'vaon12'
libva info: Trying to open D:/msys64/clangarm64/bin/\vaon12_drv_video.dll
libva info: Found init function __vaDriverInit_1_18
libva info: va_openDriver() returns 0
D:\msys64\clangarm64\bin\vainfo.exe: VA-API version: 1.18 (libva 2.18.2)
D:\msys64\clangarm64\bin\vainfo.exe: Driver version: Mesa Gallium driver 23.1.2 for D3D12 (Qualcomm(R) Adreno(TM) 680 GPU)
D:\msys64\clangarm64\bin\vainfo.exe: Supported profile and entrypoints
      VAProfileH264ConstrainedBaseline: VAEntrypointVLD
      VAProfileH264Main               : VAEntrypointVLD
      VAProfileH264High               : VAEntrypointVLD
      VAProfileHEVCMain               : VAEntrypointVLD
      VAProfileHEVCMain10             : VAEntrypointVLD
      VAProfileVP9Profile0            : VAEntrypointVLD
      VAProfileVP9Profile2            : VAEntrypointVLD
      VAProfileNone                   : VAEntrypointVideoProc
```

`VAEntrypointVLD` is decode support.

tommyvct:
Well, how about the 8cx gen 3 on the Surface Pro 9?

driver1998:
Gen 3 is decode only as well, got this output from a friend with Dev Kit 2023:

driver1998:
I guess you may already know, but ARM64 Windows support of LuaJIT [seems to be ready](https://github.com/LuaJIT/LuaJIT/issues/593), cross-compiling from Windows x64 MSVC is possible too https://github.com/LuaJIT/LuaJIT/pull/1081.

shijunz:
> Yeah, I did notice that issue have been closed.
> 
> I'm currently waiting for #186 to be merged.

@tommyvct could you double check the #186 , it maybe is  ok today now?

tommyvct:
> > Yeah, I did notice that issue have been closed.
> > 
> > I'm currently waiting for #186 to be merged.
> 
> @tommyvct could you double check the #186 , it maybe is  ok today now?

There are still something to be done on top of that PR. I have been lazy and busy, I will try to find a time to try if it works

i-lina:
Thanks @tommyvct your help in this would be much appreciated.

i-lina:
Hello @tommyvct,
Were you able to make any progress on this?
Thanks!

tommyvct:
> Hello @tommyvct, Were you able to make any progress on this? Thanks!

This PR will most likely be superseded by #190.

There are still quite a few things that we are waiting to be fixed in the upstream.

RytoEX:
Closing this since #190 has been merged.

tommyvct:
> Closing this since #190 has been merged.

Congratulations! Thank you @PatTheMav and @RytoEX  for your hard work! 

Now, I really hope someone from Qualcomm could came and start working on the encoding hardware acceleration.

trcrsired:
f Qt for arm64ec 
ARM64EC MUST DIE
https://github.com/trcrsired/Why-Microsoft-ARM64EC-ABI-MUST-DIE/blob/main/README.md

PR #143, opened 2022-11-17
macOS: Fix shared library symlinks breaking Xcode bundle composition

PatTheMav:
### Description
Fixes broken Xcode bundle composition by removing symlink chains from unversioned dynamic library names.

### Motivation and Context
CMake by default generates dynamic libraries with a full semver filename and two symlinks:

* A major version symlink pointing to the semver file
* An unversioned symlink pointing to the major version symlink

Xcode's file copy step used to move Frameworks into an app bundle only resolves a single level of symlinks, so with this setup a symlink is copied into the bundle, which then fails at the codesigning step.

Behaviour has been reported to CMake (https://gitlab.kitware.com/cmake/cmake/-/issues/24174#note_1277980) as well as Apple. These fixups should mitigate this in the short term.

### How Has This Been Tested?
Built all fixed libraries locally and checked generated symlinks.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/60-mbedtls.zsh
+++ b/deps.ffmpeg/60-mbedtls.zsh
@@ -162,12 +162,28 @@ EOF"
 fixup() {
   cd "${dir}"
 
-  if [[ ${target} == "windows-x"* ]] {
-    log_info "Fixup (%F{3}${target}%f)"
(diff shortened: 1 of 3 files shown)


RytoEX (approved):
LGTM

PR #148, opened 2022-11-28
Update obs-deps for CMake 3.0 macOS update

PatTheMav:
### Description
Fixes generated macOS libraries to not use versioned SONAMEs if possible and fix MACHO header `id` values to match those unversioned variants.

### Motivation and Context
Using versioned library names is common on Linux to discern between the _actual_ version of a library (`some_lib.so.4.5.1`), the compatibility version (usually the major version `some_lib.so.4`) that should be API-compatible, and a fallback name (`some_lib.so`), with the latter two pointing to the actual library via symlinking.

macOS uses two values in its `LC_ID_DYLIB` header (`current` and `compatibility`) which are used by its library loader to check if a library is compatible - the actual file names are not used.

A binary should link to a library (`some_lib.dylib`) via the `LC_LOAD_DYLIB` header, which itself also contains `current` and `compatibility` values, which is used by the library loader to check if a found file matches the requirements.

To that end it is more desirable to create libraries without any versioned filenames or `id`s and resolve compatibility with the header values. Alas most libraries used by OBS are designed for Linux and not macOS, so they use versioned SONAMEs to generate output names and the `install_name` given to the linker on macOS.

These changes fix not only the `rpath` entries on all libraries, but also the `id` values of libraries and local dependency to create a "clean" dependency tree on macOS.

**EDIT**: Important to note that this does _not_ change the versions of the generated libraries, just the id *name* is changed.

### How Has This Been Tested?
Built dependencies locally and compiled OBS from master tree as well as with upcoming Xcode-focused CMake update, both generated application bundles ran and resolved libraries without issue.

This should not break on master, as the current build system copies all libraries (actual library and versioned+unversioned symlinks) into the application bundle, so even with an unversioned `id`, the actual library will be found.

### Types of changes

### Checklist:

### DIFF
--- a/build-deps.zsh
+++ b/build-deps.zsh
@@ -98,6 +98,9 @@ package() {
     if [[ -d share ]] rm -rf share/^(swig)(N)
     if [[ -d cmake ]] rm -rf cmake
     if [[ -d man ]] rm -rf man
+
+    mkdir -p share/obs-deps
+    echo "${current_date}" >! share/obs-deps/VERSION
   }
 
   log_status "Create archive ${filename}"
--- a/build-deps.zsh
+++ b/build-deps.zsh
@@ -85,13 +85,13 @@ package() {
   if [[ ${PACKAGE_NAME} != 'qt'* ]] {
     log_status "Cleanup unnecessary files"
 
-    rm -rf lib/^(*.dylib|libajantv*|*.so*|*.lib)(N)
+    rm -rf lib/^(*.dylib|libajantv*|*.a|*.so*|*.lib|cmake)(N)
     rm -rf bin/^(*.exe|*.dll|*.pdb|swig)(N)
 
     if [[ -f bin/swig ]] {
       swig_lib=(share/swig/*(/))
       pushd ${swig_lib:h}
-      ln -s ${swig_lib:t} CURRENT
+      ln -sf ${swig_lib:t} CURRENT
       popd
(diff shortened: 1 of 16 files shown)


PatTheMav:
@RytoEX would like to get this merged after v29 has stabilised, is required for the macOS build system rework.

PatTheMav:
@RytoEX branch represents RC status now, also renamed the PR as discussed.

RytoEX:
Key questions from my point of view:
1. Do the resultant deps still work with the current obs-studio CMake (the CMake 2.0 Build System)?
2. Do the resultant deps work with the proposed next-gen CMake (the CMake 3.0 Build System)?

PatTheMav:
> Key questions from my point of view:
> 
> 1. Do the resultant deps still work with the current obs-studio CMake (the CMake 2.0 Build System)?
> 2. Do the resultant deps work with the proposed next-gen CMake (the CMake 3.0 Build System)?

It is _required_ for CMake 3.0 to work at all (resulting app bundle will not work without it) and in my tests with the current build system the resulting app bundle ran just fine.

I _did_ find an issue with SDL2 being bundled with FFmpeg again (the formula wasn't correctly hidden by the build script), but it seems there is a new official switch to disable SDL2 which fixed that.

RytoEX (approved):
Looks good to me.

PR #155, opened 2023-02-02
Add obs-websocket headers-only libraries

tytan652:
### Description

Adds to obs-deps headers-only deps from obs-websocket:
- nlohmann JSON 3.11.2 (latest) and its cmake package.
- WebSocket++ 0.8.2 (latest) same version as obs-websocket submodule.
  - This library has a CMake package but its version check is broken, so a custom finder was made for obs-studio.
- Asio 1.12.1 same version as obs-websocket submodule.
  - Unlike the other two, Asio is installed by copying it's headers directly in the include directory to mainly avoid depending on autotools and GNU Make. Also test builds (through autoconf+make) without Boost does not seem to work in this version.
  - A finder was made for obs-studio.
  
Also the build.deps change is not the same as in #148, so normally those two PR should not conflict.

### Motivation and Context

Reduce the number of submodules in obs-studio repo and provides those libraries in obs-deps.

### How Has This Been Tested?

obs-studio CI on my fork (https://github.com/tytan652/obs-studio/pull/10) with obs-websocket modified to be aware of obs-deps and custom-made finders for Asio and WebSocket++.

### Checklist:


deps.macos/80-asio.zsh
```diff
@@ -0,0 +1,31 @@
...
+
+install() {
+  autoload -Uz progress
+
+  log_info "Install (%F{3}${target}%f)"
+  cd "${dir}/asio"
+
+  mkdir -p ${target_config[output_dir]}/include
+
+  local -a options=("" "-R")
```
PatTheMav, line 23: I would use a simple globbing expression here:

`cp -R include/(asio|asio.hpp) "${target_config[output_dir]}/include"`

PatTheMav:
Except for the globbing change in the `asio` build script for macOS, this looks fine to me.

Are there any expectations/requirements for any of these to be built as a shared library vs static (I'm trying to default to static libs on macOS if the corresponding library is used by a single obs plugin only, with freetype being an exception as it needs to have `libpng` linked statically already).

tytan652:
> Are there any expectations/requirements for any of these to be built as a shared library vs static (I'm trying to default to static libs on macOS if the corresponding library is used by a single obs plugin only, with freetype being an exception as it needs to have `libpng` linked statically already).

Those are headers-only library, There is no library file (.so,.a,.dylib,.lib,.dll). Static or shared it change nothing.

deps.macos/80-asio.zsh
tytan652: And I log what with the cp command ?

tytan652: I mostly made it this way to be similar to its Windows counterpart.

tytan652: Made changes.

deps.macos/80-asio.zsh
```diff
@@ -0,0 +1,25 @@
...
+
+install() {
+  autoload -Uz progress
+
+  log_info "Install (%F{3}${target}%f)"
+  cd "${dir}/asio"
+
+  mkdir -p ${target_config[output_dir]}/include
+
+  log_output "include => ${target_config[output_dir]}/include"
```
PatTheMav, line 23: I would move this below the copy command and have it say `log_status "Copied headers to ${target_config[output_dir]}"`.

Usually I went with `log_debug` before doing things and `log_status` after it was successful.

tytan652: Made the change.

PatTheMav (approved):
Looks good to me.

tytan652:
- Tried to add explanation for Asio 1.12.1
- Patched WebSocket++ to not break if a future version of CMake remove compatibility with 2.8.8
  - CMake emits a deprecation warning about dropping it in a future version

PatTheMav:
@tt2468 any time to ponder this PR since our last call?

tt2468:
This looks good to me. Only thing I ask for is a brief description of what will be required to migrate an existing build environment on a given platform to this new one, so that we can post it in the inevitable tag description.

tytan652:
> This looks good to me. Only thing I ask for is a brief description of what will be required to migrate an existing build environment on a given platform to this new one, so that we can post it in the inevitable tag description.

Windows and macOS: Switch to latest obs-deps
Ubuntu: Re-run dependency install script
Flatpak: Use the latest package

Bonus: Ask developers to use the headers-only libraries provided by obs-deps or the distro/Flatpak.

RytoEX (approved):
Looks good to me.

PR #157, opened 2023-03-02
deps.windows: Add ducible

derrod:
### Description

Adds the [ducible](https://github.com/jasonwhite/ducible) tool to windows deps.

Unfortunately ducible does not use CMake, but we don't need to link against it and just want the command line tools. This PR includes patches to make the included project work with VS 2022 and to fix a potential issue with invocation of the generic `python` command.

### Motivation and Context

For https://github.com/obsproject/obs-studio/pull/8220 to work at its best, we need to overwrite/remove certain parts from the PE metadata (e.g. timestamps). Ducible handles this and can be relatively easily be executed as part of CMake's custom post-build commands system.

Instead of relying on a third-party repository and code downloads from there, this compiles the tools on our CI and makes them part of obs-deps so they can be run as part of a CMake post-build command locally or on CI when creating reproducible builds.

### How Has This Been Tested?

Built locally.

### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/70-ducible.ps1
+++ b/deps.windows/70-ducible.ps1
@@ -1,0 +1,109 @@
+param(
+    [string] $Name = 'ducible',
+    [string] $Version = 'v1.2.2',
+    [string] $Uri = 'https://github.com/jasonwhite/ducible.git',
+    [string] $Hash = 'b7810415529f801d98203cba00db969a6ef88a14',
+    [array] $Patches = @(
+        @{
+            PatchFile = "${PSScriptRoot}/patches/ducible/0001-Remove-8.1-target-platform.patch"
+            HashSum = 'f45851456fd93e4491e42f108ff19e60c0de3df35f1141208c95623edae14517'
+        }
+        @{
+            PatchFile = "${PSScriptRoot}/patches/ducible/0002-Use-py-instead-of-python.patch"
+            HashSum = '5922aa7639b2a364c0f90541dc15c05dc804ace9178a40a661d360bb961584c4'
+        }
+    )
+)
+
+function Setup {
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
+}
+
+function Patch {
+    Log-Information "Patch (${Target})"
+    Set-Location $Path
+
+    $Patches | ForEach-Object {
+        $Params = $_
+        Safe-Patch @Params
+    }
+}
+
+function Clean {
+    Set-Location $Path
+
+    if ( Test-Path "build_${Target}" ) {
+        Log-Information "Clean build directory (${Target})"
+        Remove-Item -Path "build_${Target}" -Recurse -Force
+    }
+}
+
+function Build {
+    Log-Information "Build (${Target})"
+    Set-Location $Path
+
+    if ( $Configuration -match "Debug" ) {
+        $DucibleConfig = 'Debug'
+    } else {
+        $DucibleConfig = 'Release'
+    }
+
+    if ( $Target -match "x64" ) {
+        $DucibleTarget = 'x64'
+    } else {
+        $DucibleTarget = 'Win32'
+    }
+
+    $Options = @(
+        '/t:Build'
+        "/p:OutDir=.\..\..\..\build_${Target}\"
+        "/p:Configuration=${DucibleConfig}"
+        "/p:Platform=${DucibleTarget}"
+        '/p:PlatformToolset=v143'
+    )
+    # Find MSBuild command
+    $MSBUILD = &"${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -prerelease -products * -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe
+    
+    Invoke-External $MSBUILD -m vs\vs2015\ducible.sln @Options
+}
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location $Path
+    
+    $Params = @{
+        ErrorAction = "SilentlyContinue"
+        Path = @(
+            "$($ConfigData.OutputPath)/bin"
+        )
+        ItemType = "Directory"
+        Force = $true
+    }
+
+    New-Item @Params *> $null
+
+    $Items = @(
+        @{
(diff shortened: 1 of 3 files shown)


deps.windows/70-ducible.ps1
```diff
@@ -0,0 +1,109 @@
...
+
+    $Options = @(
+        '/t:Build'
+        "/p:OutDir=.\..\..\..\build_${Target}\"
+        "/p:Configuration=${DucibleConfig}"
+        "/p:Platform=${DucibleTarget}"
+        '/p:PlatformToolset=v143'
+    )
+    # Find MSBuild command
+    $MSBUILD = &"${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -prerelease -products * -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe
```
PatTheMav, line 65: Use `Invoke-Devshell` instead: 

https://github.com/obsproject/obs-deps/blob/9d4d2fdc5b1f840d74a483e3c61aa15163079366/deps.windows/40-detours.ps1#L40

With that you should just be able to pass `msbuild -m <SOLUTION> @Options` as the `BuildCommand`.

deps.windows/patches/ducible/0002-Use-py-instead-of-python.patch
```diff
@@ -0,0 +1,93 @@
...
+diff --git a/vs/vs2015/ducible/ducible.vcxproj b/vs/vs2015/ducible/ducible.vcxproj
+index 99d4ef4..f49dd22 100644
+--- a/vs/vs2015/ducible/ducible.vcxproj
++++ b/vs/vs2015/ducible/ducible.vcxproj
+@@ -109,7 +108,7 @@
+       <GenerateDebugInformation>true</GenerateDebugInformation>
+     </Link>
+     <CustomBuildStep>
+-      <Command>python ..\..\..\scripts\version.py ..\..\..\src\version.h.in ..\..\..\src\version.h --version-file ..\..\..\VERSION</Command>
++      <Command>py -3 ..\..\..\scripts\version.py ..\..\..\src\version.h.in ..\..\..\src\version.h --version-file ..\..\..\VERSION</Command>
```
PatTheMav, line 20: Does this require specific steps to work? Will Python 3 always be available as `py3` in the path?

If we patch this, it needs to work universally and can't just be patched to work on CI.

deps.windows/70-ducible.ps1
derrod: How did I miss that?

deps.windows/patches/ducible/0002-Use-py-instead-of-python.patch
derrod: `py` is the "Python Launcher for Windows" and is installed by default when python is installed via their official Windows installers: https://docs.python.org/3/using/windows.html#launcher

`py -3` in this case runs the script with the latest Python 3.x version installed on the system. Likewise `py -3.10` can be used to run a specific version.

deps.windows/70-ducible.ps1
derrod: Unfortunately, running the build command from within the devshell seems to mess up the environment so that external tools are no longer found when running `msbuild`. Also happens when running the dev shell manually. Confusingly, `py` and `python` still work when inside the shell, but not inside `msbuild` anymore?!

deps.windows/patches/ducible/0002-Use-py-instead-of-python.patch
PatTheMav: So the "Python Launcher for Windows" becomes a Windows build dependency if reproducible builds are desired?

deps.windows/70-ducible.ps1
PatTheMav: So it doesn't even work when run in a normal dev shell either?

deps.windows/patches/ducible/0002-Use-py-instead-of-python.patch
derrod: obs-deps itself is not reproducible right now, and that will take more time. The main reason of including ducible in obs-deps would be to avoid having to download it from *some other source* on CI or during local builds that are meant to be identical to CI builds (i.e. using the prebuilt obs-deps).

However, due to the difficulties encountered so far I am considering shelving ducible for now and then come back to it post-30 with a fork with CMake support. Just to make everyone's life a little easier.

deps.windows/70-ducible.ps1
derrod: If I launch the dev shell via the powershell script included with VS (`Launch-VsDevShell.ps1`) it does work. However, launching the dev shell in the same way the helper script in obs-deps does does not.

RytoEX:
I was going to ask if you would you consider upstreaming patches to ducible to give them a working VS2022 project or CMake support, but I see the project has been inactive since 2019. Are there any other comparable projects that are active?

derrod:
> I was going to ask if you would you consider upstreaming patches to ducible to give them a working VS2022 project or CMake support, but I see the project has been inactive since 2019. Are there any other comparable projects that are active?

I'm not aware of any, and given that the PE format probably isn't going to change ever it doesn't really need one. Perhaps I can try upstreaming CMake at least if I get that going.

deps.windows/70-ducible.ps1
PatTheMav: So here's how to make it work:

* Make it depend on 60-python
* Add the python from pyenv to the path
* Run `Invoke-DevShell`
  * I used `Invoke-DevShell -BasePath $(Get-Location | Convert-Path) -BuildPath . -BuildCommand $BuildCommand -Target x64` with `msbuild -m vs\vs2015\ducible.sln /t:Build /p:Configuration=Debug /p:Platform=x64 /p:PlatformToolset=v143` as the build command
* Profit

derrod:
Adding this tool was supposed to help with delta size and reproducibility by setting various metadata fields to known values, however recent MSVC versions do this with the `/Brepro` flag to a satisfying extent, and codesigning the file with `signtool` actually re-sets some of those metadata fields to their "correct" values anyway, thus making this kind of pointless.

PR #159, opened 2023-03-02
deps.windows: Add VLC to Windows dependencies

PatTheMav:
### Description
Adds libVLC headers to obs-deps.

### Motivation and Context
Reduces complexity for build system setup on Windows as obs-deps will contain almost every build-time dependency. Also limits possible availability issues of videolan's servers to the time of an obs-deps release instead of every time obs-studio is built.

### How Has This Been Tested?
Build obs-deps with both dependencies, checked headers are put to the correct destinations.

### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/90-vlc.ps1
+++ b/deps.windows/90-vlc.ps1
@@ -1,0 +1,54 @@
+param(
+    [string] $Name = 'vlc',
+    [string] $Version = '3.0.8',
+    [string] $Uri = 'https://cdn-fastly.obsproject.com/downloads/vlc.zip',
+    [string] $Hash = "${PSScriptRoot}/checksums/vlc.zip.sha256"
+)
+
+function Setup {
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
+}
+
+function Build {
+    Log-Information "Build (${Target})"
+    Set-Location $Path
+
+    $VersionString = [System.Version] $Version
+
+    (Get-Content -Path include/vlc/libvlc_version.h.in -Raw) `
+        -replace "@VERSION_MAJOR@", "$($VersionString.Major)" `
+        -replace "@VERSION_MINOR@", "$($VersionString.Minor)" `
+        -replace "@VERSION_REVISION@", "$($VersionString.Build)" `
+        | Out-File -Path include/vlc/libvlc_version.h -NoNewLine
+}
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location $Path
+
+    $Params = @{
+        ErrorAction = "SilentlyContinue"
+        Path = @(
+            "$($ConfigData.OutputPath)/include"
+        )
+        ItemType = "Directory"
+        Force = $true
+    }
+
+    New-Item @Params *> $null
+
+    $Items = @(
+        @{
+            Path = "include/vlc"
+            Destination = "$($ConfigData.OutputPath)/include/"
+            Recurse = $true
+            Force = $true
+        }
+    )
+
+    $Items | ForEach-Object {
+        $Item = $_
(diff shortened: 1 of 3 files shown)


deps.windows/90-vlc.ps1
```diff
@@ -0,0 +1,54 @@
+param(
+    [string] $Name = 'vlc',
+    [string] $Version = '3.0.8',
+    [string] $Uri = 'https://cdn-fastly.obsproject.com/downloads/vlc.zip',
+    [string] $Hash = "${PSScriptRoot}/checksums/vlc.zip.sha256"
```
RytoEX, line 5: We actually use the git repo on Windows, because we wanted the latest headers. If nothing else, we should be using latest stable 3.x.

deps.windows/checksums/vlc.zip.sha256
```diff
@@ -0,0 +1,14 @@
...
+  <Obj RefId="0">
+    <TN RefId="0">
+      <T>Microsoft.PowerShell.Commands.FileHashInfo</T>
+      <T>System.Object</T>
+    </TN>
+    <ToString>Microsoft.PowerShell.Commands.FileHashInfo</ToString>
+    <Props>
+      <S N="Algorithm">SHA256</S>
+      <S N="Hash">91F589EF69FCE51645A3ECBB215B405C98DB7B891479474EC3B5ED3B63C25E4A</S>
+      <S N="Path">windows_32-bit_build_temp\vlc.zip</S>
```
RytoEX, line 11: Why 32-bit?

RytoEX (changes requested):

deps.windows/90-vlc.ps1
PatTheMav: You mean the contents of the zip are actually just a copy of the git repo?

deps.windows/checksums/vlc.zip.sha256
PatTheMav: I just copied another checksum file - the Path is never used and could in theory even be omitted.

deps.windows/90-vlc.ps1
RytoEX: Oh. I have no idea what vlc.zip on fastly actually contains, now that you mention it. For some reason, I thought we just did a git checkout on CI. Inspecting it, it does appear to be some zip of the VLC source, and I'm not sure which version. We just mark it as 3.0.0-git on CI.

PatTheMav: We _could_ just use a git checkout of 3.0.8 then?

RytoEX: For now, let's just do what this PR does since that's what obs-studio CI does. We can change it later to either do a git checkout or a source download of some kind. If that becomes unreliable in the future for some reason (server outages), we can make a mirror again then.

deps.windows/checksums/vlc.zip.sha256
RytoEX: Path is actually used here:
* https://github.com/obsproject/obs-deps/blob/713379c1cd2f556ca2c759c0e139a985273c0ff7/utils.pwsh/Invoke-SafeWebRequest.ps1#L76-L80

PatTheMav: Yeah I guess I limited it to the filename component because of the 32bit/64bit dichotomy and I didn't want to maintain different checksums for each architecture.

Would be better to output just the filenames and not full paths then.

RytoEX: Discussed off-thread, and this should work as-is, but we'll likely remove all leading path components from the Windows hash files in a separate PR after.

RytoEX (approved):

PR #166, opened 2023-03-17
deps.ffmpeg: Fix AOM target CPU arch

RytoEX:
### Description

AOM allows you to set the target CPU arch (AOM_TARGET_CPU) at configure time. If not specified, their CMake will attempt to autodetect this value based on CMAKE_SYSTEM_PROCESSOR. The value of AOM_TARGET_CPU is then used to enable various optimizations.

We used to not set a value for AOM_TARGET_CPU.[1] We started setting a value for it with the unified FFmpeg build. However, we set the value of "arch" based on the build target string (i.e., windows-x86 or windows-x64), which results in an "arch" of "x86" or "x64". AOM's CMake does expect the value "x64" as a possible value of AOM_TARGET_CPU, as it instead expects "x86_64".[2] However, it will not produce a configure error if given "x64" for the value of AOM_TARGET_CPU. It will accept this value, but then subsequent checks will rely on the value being "x86" or "x86_64" for things like using YASM or NASM.

AOM's CMake will correctly set this from the value of CMAKE_SYSTEM_PROCESSOR, which we already set for Windows builds, so let's just remove AOM_TARGET_CPU as a global override option, and instead also specify CMAKE_TOOLCHAIN_FILE for macOS builds.

References:
[1]: https://github.com/obsproject/obs-deps/blob/6366c22b731c31112449d1d323daafaf8e306e3c/CI/windows/build_libaom.sh#L28-L43 [2]: https://aomedia.googlesource.com/aom.git/+/refs/tags/v3.6.0/build/cmake/aom_configure.cmake#57
 *   https://cmake.org/cmake/help/v3.26/variable/CMAKE_SYSTEM_PROCESSOR.html
 *   https://aomedia.googlesource.com/aom.git/+/refs/tags/v3.4.0/build/cmake/aom_configure.cmake#53

### Motivation and Context

Want optimizations when available.

### How Has This Been Tested?

I've built AOM locally both in x86 and x86_64 (x64) configurations and built the whole FFmpeg dependency chain this way. You can see the change reflected in the build log:
```
--- aom_configure: Detected CPU: x86_64
```
where it used to say:
```
--- aom_configure: Detected CPU: x64
```

See also:
* https://github.com/obsproject/obs-deps/actions/runs/4448056886/jobs/7810352225#step:6:1286
* https://github.com/obsproject/obs-deps/actions/runs/4448056886/jobs/7810352547#step:6:2278

I have not yet tested the subsequent build in OBS, but I don't suspect anything will break. If something does, we can easily revert this.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/40-aom.zsh
+++ b/deps.ffmpeg/40-aom.zsh
@@ -55,7 +55,6 @@ config() {
   args=(
     ${cmake_flags}
     -DBUILD_SHARED_LIBS="${_onoff[(( shared_libs + 1 ))]}"
-    -DAOM_TARGET_CPU="${arch}"
     -DENABLE_DOCS=OFF
     -DENABLE_EXAMPLES=OFF
     -DENABLE_TESTDATA=OFF
@@ -65,7 +64,7 @@ config() {
   )
 
   case ${target} {
-    macos-arm64) args+=(-DCONFIG_RUNTIME_CPU_DETECT=0) ;;
+    macos-*) args+=(-DCONFIG_RUNTIME_CPU_DETECT=0 -DCMAKE_TOOLCHAIN_FILE="build/cmake/toolchains/${target_config[cmake_arch]}-macos.cmake") ;;
     windows-x*) args+=(-DCMAKE_TOOLCHAIN_FILE="build/cmake/toolchains/${target_config[cmake_arch]}-mingw-gcc.cmake")
   }
 


deps.ffmpeg/40-aom.zsh
```diff
@@ -55,7 +55,6 @@ config() {
   args=(
     ${cmake_flags}
     -DBUILD_SHARED_LIBS="${_onoff[(( shared_libs + 1 ))]}"
-    -DAOM_TARGET_CPU="${arch}"
```
tytan652, line 58: Maybe this would work.
```sh
 case ${target} {
    macos-arm64) args+=(-DCONFIG_RUNTIME_CPU_DETECT=0 -DAOM_TARGET_CPU="${arch}") ;;
    windows-x*) args+=(-DCMAKE_TOOLCHAIN_FILE="build/cmake/toolchains/${target_config[cmake_arch]}-mingw-gcc.cmake")
  }
```

PR #170, opened 2023-03-25
deps.ffmpeg: Add libdatachannel

DDRBoxman:
### Description

Adds libdatachannel to the ffmpeg deps allowing us to use webrtc in OBS.

Libdatachannel requires srtp to be enabled in mbedtls so we enable that here.

libdatachannel is compiled in the ffmpeg deps to cut down on complexity and due to the fact that there are some lingering issues with mbedtls combined with  libdatachannel with the msvc compiler.

### Motivation and Context

This will allow us to add WHIP, WHEP, and eventually maybe even build out webrtc apis for plugins.

Prerequisite for this PR:
https://github.com/obsproject/obs-studio/pull/7926

### Types of changes

### Checklist:


pkviet:
@DDRBoxman 
Here's a patch which does the following:
- update mbedtls to 3.4.0
- update datachannel to the 0.19.0 tag
- fix the cmake files generated at install to be useable by the ps1 scripts (the cmake file refer to .dll.a import libs which work when cross-compiling; but for msvc we need the .lib files generated at fixup)
When the datachannel ps1 script is run with 0.19.0 tag the compilation works and the dll loads without crashing in obs.
However the whip connection fails after a while; the whep player doesn't play anything.
[obs-deps-pkv_changes.patch](https://github.com/obsproject/obs-deps/files/11137856/obs-deps-pkv_changes.patch)
This other patch is for the datachannel ps1:
[obs-deps-pkv2_changes.patch](https://github.com/obsproject/obs-deps/files/11137884/obs-deps-pkv2_changes.patch)

deps.ffmpeg/60-mbedtls.zsh
```diff
@@ -3,21 +3,27 @@ autoload -Uz log_debug log_error log_info log_status log_output
 ## Dependency Information
 local name='mbedtls'
 local -A versions=(
-  macos 3.2.1
-  linux 3.2.1
-  windows 3.2.1
+  macos 3.4.0
+  linux 3.4.0
+  windows 3.4.0
```
RytoEX, line 8: Did we determine in the end which version of mbedTLS is safe to use for our needs?

deps.ffmpeg/70-libdatachannel.zsh
```diff
@@ -0,0 +1,112 @@
+autoload -Uz log_debug log_error log_info log_status log_output
+
+## Dependency Information
+local name='libdatachannel'
+local version='0.19.0'
+local url='https://github.com/paullouisageneau/libdatachannel.git'
+local hash='004c6b74ffa1d2718499b84d5bd7674d94b83dbc'
```
RytoEX, line 7: Are there any newer commits that we'd like to use or is this the latest we're comfortable using?

deps.ffmpeg/70-libdatachannel.zsh
```diff
@@ -0,0 +1,112 @@
...
+fixup() {
+  cd "${dir}"
+
+  case ${target} {
+    windows*)
+      log_info "Fixup (%F{3}${target}%f)"
+      if (( shared_libs )) {
+        autoload -Uz create_importlibs
+        create_importlibs ${target_config[output_dir]}/bin/libdatachannel*.dll
+        log_status "Fixing cmake for msvc import library syntax"
```
RytoEX, line 99: Nits:
* cmake -> CMake
* msvc -> MSVC

RytoEX (changes requested):
Squash the second commit since it's just reverting changes in the first one.

@PatTheMav do we want to split this into individual commits to update+patch mbedTLS, add libdatachannel (or update mbedTLS, add mbedTLS patch and libdatachannel)? I usually prefer component updates are at least their own commits. Mostly unsure how to treat the added patch.

deps.ffmpeg/60-mbedtls.zsh
pkviet: I personally tested with 3.4.0 which does work.
Maybe 3.2.1 would work too though.

deps.ffmpeg/70-libdatachannel.zsh
Sean-Der: This is the latest commit/one we want

deps.ffmpeg/60-mbedtls.zsh
Sean-Der: 3.4.0 is what we want. mbedTLS fixed their includes on Windows with 3.4.0

Sean-Der:
@RytoEX mind reviewing again? I split the PR into three commits and addressed all other comments.

* Upgrade mbedTLS
* Enable DTLS-SRTP with patch in mbedTLS
* Add libdatachannel

deps.ffmpeg/70-libdatachannel.zsh
Sean-Der: Fixed!

deps.ffmpeg/60-mbedtls.zsh
RytoEX: > mbedTLS fixed their includes on Windows with 3.4.0

Not that I'm aware of. The GitHub Issue cited to me was:
https://github.com/Mbed-TLS/mbedtls/issues/7087

deps.ffmpeg/70-libdatachannel.zsh
RytoEX: It's not the latest commit: https://github.com/paullouisageneau/libdatachannel/compare/004c6b74ffa1d2718499b84d5bd7674d94b83dbc...0abde70beb1310bd393af6cddd6d1a695fe88131

If it's "the one we want for now" because it's what was tested against, then that's fine. We'll likely update it before the next major/minor release of OBS anyway, as we update dependencies when possible for major/minor releases of OBS.

Sean-Der: @RytoEX I am not sure what to respond with here? This is the commit I would prefer to use.

deps.ffmpeg/60-mbedtls.zsh
Sean-Der: @RytoEX I believe the particular header causing us issues was fixed. Looking at the discussion I believe others are struggling with different includes.

https://discord.com/channels/348973006581923840/1052657814264356955/1092873798749016155 is the relevant conversation in development channel.

deps.ffmpeg/70-libdatachannel.zsh
RytoEX: I assume it's simply both of us misunderstanding the other. I intended to ask, "Are there any newer commits that we'd like to use? If not, is it because this is the latest commit we're comfortable using?"

I read your answer as, "This is the latest commit or the one we want," so I rebutted the idea that it's the latest commit, and then attempted to clarify my position that if the reason this commit is chosen is simply "because it's the one we want", then that's fine.

Sean-Der: Sorry my misunderstanding. 

This is the commit we want. I haven't audited/checked the commits that have arrived after this. I am not comfortable including other commits.

RytoEX: Then we'll keep this commit for now. As mentioned, we'll likely update deps if they have updates available when we prep the next deps update for the next major/minor release of OBS Studio.

deps.ffmpeg/60-mbedtls.zsh
RytoEX: The linked GitHub Issue is about Designated Initializers (a GCC C++ extension not in the C++ standard until C++20) in header files. My replies in that conversation were not refuted directly, so I had no reason to believe that the issue with Designated Initializers was resolved intentionally.

My relevant replies in the conversation that you linked:
> IIRC, the issue is that their headers use Designated Initializers, which work in GCC because of GCC extensions, but do not work in standard C++ in MSVC until C++20, and they aren't comfortable using C++20 as their baseline yet.
> I don't think we tested with 3.4.0 to find out. It wasn't out yet.
> 3.4.0 only came out last week (March 28).
> So we wouldn't have tested it two weeks ago.
> If they accidentally fixed it, neat. They don't seem to be aware of that being the case though, or they would have closed the Issue.

If, in fact, the GitHub Issue cited to me was not the actual problem, which seems to be implied by [your follow-up comment](https://github.com/paullouisageneau/libdatachannel/pull/822#issuecomment-1478892712) on the libdatachannel PR to add mbedTLS support, then it was simply not something I recalled or perhaps not clarified in conversations that I followed closely.

Sean-Der: @RytoEX My comment is not related to this. The required includes changed. mbedTLS 3.4.0 wasn’t working on all platforms.

I am pretty confident that code mbedTLS stopped usage of designated initializations.

what do we need here to move forward?

@pkviet does what I say match with what you saw?

Sean-Der: @RytoEX I will fix the Windows builds now.

Sean-Der:
@RytoEX mind reviewing again? I enabled POSIX with mingw since libdatachannel depends on it.

RytoEX:
> @RytoEX mind reviewing again? I enabled POSIX with mingw since libdatachannel depends on it for c++11 threading.

I've asked @PatTheMav to assist in review. From what I recall of previous conversations, there should have been a successful build without `-posix` before (@pkviet tested it), and I'm not keen on just switching to `-posix` for our entire cross-compile build without at least some investigation.

PatTheMav:
A bit of background on the whole thing:

* x86_64-w64-mingw32-g++-posix is an implementation of C++11's native threading features _using_ GCC's gthreads/pthreads - this automatically adds MingW-w64's pthreads library (`libwinpthread-1.dll`) as a dependency for any generated DLL.
* We default to `x86_64-w64-mingw32-g++-win32` (the default alias for `x86_64-w64-mingw32-g++`) because we do not wish to distribute any MingW-w64 DLLs and ensure that native Win32-API features are used.

There are some complications around this:
* `aom` and `libvpx` have built-in shims that emulate native `pthread` support using Win32 threads (no MingW-w64 trickery necessary)
* For `mbedTLS` we effectively disable thread support on Windows to avoid any issues while cross-compiling
* `srt` was the most troublesome of the whole bunch - it either has a hard requirement for `pthreads` (no matter the platform) _or_ can use C++11 threads instead (provided natively by MSVC's toolchain). When cross-compiling this creates an obvious issue:
    * The created DLL will either directly or indirectly require `winpthread` to be shipped with them
* `librist` also has a `pthread` shim that could be used - I don't remember the exact reason but I recall there were issues with the shim's symbols clashing with other libraries when compiling FFmpeg, which is why we elected not to use it anymore
* FFmpeg supports `pthreads` as well as Win32 threads, we chose each based on the following criteria:
    * When compiling FFmpeg with _shared_ 3rd party libraries, we use Win32 threads - any possible pthreads-based symbols (shims or native) don't leak outside of the DLLs and won't interfere with FFmpeg
    * When compiling FFmpeg with _static_ 3rd party libraries, we use `pthreads` as some of the 3rd party library carry `pthreads` as a private usage requirement anyway - FFmpeg will link against `pthread` and we have to force `MingW-w64` to use static linking (and just like with `srt` we have to hide all DLLs in the filesystem to make the compiler actually use static linking)

Given that libdatachannel will not be used by FFmpeg (and is just compiled as part of it as a workaround for the time being), it should be ok to use the `-posix` variants to enable use of C++11 threads in it. As far as I can tell the compile flags enforce static linking of pthreads (just like with `srt`) and also creating a DLL so its own `pthreads` symbols don't clash with `libobs` own `w32-pthreads` variant.

deps.ffmpeg/60-mbedtls.zsh
```diff
@@ -3,21 +3,27 @@ autoload -Uz log_debug log_error log_info log_status log_output
...
-  linux 869298bffeea13b205343361b7a7daf2b210e33d
-  windows 869298bffeea13b205343361b7a7daf2b210e33d
+  macos 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
+  linux 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
+  windows 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
 )
 local -a patches=(
   "macos ${0:a:h}/patches/mbedtls/0001-enable-posix-threading-support.patch \
     ea52cf47ca01211cbadf03c0493986e8d4e0d1e9ab4aaa42365b2dea7b591188"
+  "macos ${0:a:h}/patches/mbedtls/0002-enable-dtls-srtp-support.patch \
```
PatTheMav, line 19: You can use a single catch-all entry for all three platforms using `*` as the platform identifier.

This requires the check on line 62 to be changed to the following:

`if [[ "${target%%-*}" == ${~_target} ]] apply_patch "${_url}" "${_hash}"`

This effectively tells `zsh` to interpret the string stored in `_target` to be interpreted as a glob expression, which makes `*` match any possible target identifier.

deps.ffmpeg/70-libdatachannel.zsh
```diff
@@ -0,0 +1,112 @@
...
+    -DNO_TESTS=1
+    -DNO_EXAMPLES=1
+  )
+  case ${target} {
+    windows-x*)
+      args+=(
+        -DCMAKE_CXX_FLAGS="-static-libgcc -static-libstdc++ -w -pipe -fno-semantic-interposition -static"
+        -DCMAKE_C_FLAGS="-static-libgcc -w -pipe -fno-semantic-interposition -static"
+        -DCMAKE_SHARED_LINKER_FLAGS="-static-libgcc -static-libstdc++ -L${target_config[output_dir]}/lib -Wl,--exclude-libs,ALL -static"
+      )
```
PatTheMav, line 48: Add a toolchain override here:

`-DCMAKE_TOOLCHAIN_FILE="${target_config[posix_toolchain]}"`.

utils.zsh/toolchain/windows-cross-toolchain.cmake
```diff
@@ -6,8 +6,8 @@ elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86")
     set(_CROSS_PREFIX "i686")
 endif()
 
-set(CMAKE_C_COMPILER ${_CROSS_PREFIX}-w64-mingw32-gcc)
-set(CMAKE_CXX_COMPILER ${_CROSS_PREFIX}-w64-mingw32-g++)
+set(CMAKE_C_COMPILER ${_CROSS_PREFIX}-w64-mingw32-gcc-posix)
```
PatTheMav, line 9: Instead of changing this globally, create a new file `windows-cross-toolchain-posix.cmake` with them.

deps.ffmpeg/70-libdatachannel.zsh
PatTheMav: Then, to enable this, add

`[posix_toolchain]="${funcsourcetrace[1]:A:h}/toolchain/windows-cross-toolchain-posix.cmake"`

to lines 49 and 57 in `30-build.zsh` to add it to the target config array.

PatTheMav:
You also need to upstream a patch to `libdatachannel`:

The `mbedtls` finder uses `objdump` to find the `SONAME` of the library - this doesn't work with toolchains: Instead the finder needs to use `CMAKE_OBJDUMP` (which should usually contain the `objdump` binary of the selected toolchain) and the toolchain files need to be updated to use the variant provided by MingW-w64.

Sean-Der:
@PatTheMav Opened a PR against libdatachannel https://github.com/paullouisageneau/libdatachannel/pull/889

deps.ffmpeg/70-libdatachannel.zsh
```diff
@@ -0,0 +1,112 @@
+autoload -Uz log_debug log_error log_info log_status log_output
+
+## Dependency Information
+local name='libdatachannel'
+local version='0.19.0'
+local url='https://github.com/paullouisageneau/libdatachannel.git'
+local hash='004c6b74ffa1d2718499b84d5bd7674d94b83dbc'
+
+## Dependency Overrides
+local -i shared_libs=1
```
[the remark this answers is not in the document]
Sean-Der: @DDRBoxman Do you remember why this was done? I am not sure

utils.zsh/toolchain/windows-cross-toolchain.cmake
Sean-Der: @PatTheMav Done. Does that look right what I did?

deps.ffmpeg/70-libdatachannel.zsh
Sean-Der: Done

deps.ffmpeg/60-mbedtls.zsh
Sean-Der: Done

utils.zsh/toolchain/windows-cross-toolchain-posix.cmake
```diff
@@ -0,0 +1,17 @@
...
+    set(_CROSS_PREFIX "x86_64")
+elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86")
+    set(_CROSS_PREFIX "i686")
+endif()
+
+set(CMAKE_C_COMPILER ${_CROSS_PREFIX}-w64-mingw32-gcc-posix)
+set(CMAKE_CXX_COMPILER ${_CROSS_PREFIX}-w64-mingw32-g++-posix)
+set(CMAKE_RC_COMPILER ${_CROSS_PREFIX}-w64-mingw32-windres)
+set(CMAKE_RANLIB ${_CROSS_PREFIX}-w64-mingw32-ranlib)
+
```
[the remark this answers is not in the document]
Sean-Der: @PatTheMav do you know what the binary name is? Sorry I don’t have a Windows install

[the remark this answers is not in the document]
Sean-Der: @PatTheMav do you know what the binary name is? Sorry I don’t have a Windows install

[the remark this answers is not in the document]
Sean-Der: @PatTheMav Ok got the values from pkv! does that look ok?

deps.ffmpeg/60-mbedtls.zsh
```diff
@@ -3,21 +3,21 @@ autoload -Uz log_debug log_error log_info log_status log_output
...
+  macos 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
+  linux 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
+  windows 1873d3bfc2da771672bd8e7e8f41f57e0af77f33
 )
 local -a patches=(
   "macos ${0:a:h}/patches/mbedtls/0001-enable-posix-threading-support.patch \
     ea52cf47ca01211cbadf03c0493986e8d4e0d1e9ab4aaa42365b2dea7b591188"
-  "linux ${0:a:h}/patches/mbedtls/0001-enable-posix-threading-support.patch \
-    ea52cf47ca01211cbadf03c0493986e8d4e0d1e9ab4aaa42365b2dea7b591188"
+  "* ${0:a:h}/patches/mbedtls/0002-enable-dtls-srtp-support.patch \
```
PatTheMav, line 19: You seem to have removed the patch enabling Posix threads on Linux - please leave this in and just add the new patch.

utils.zsh/toolchain/windows-cross-toolchain-posix.cmake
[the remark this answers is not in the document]
PatTheMav: Looks correct to me.

deps.ffmpeg/60-mbedtls.zsh
Sean-Der: @PatTheMav fixed! Sorry about that

Sean-Der:
@PatTheMav the patch has been upstreamed and merged. Could I get another review please and the workflow approved. Will debug if anything fails!

deps.ffmpeg/60-mbedtls.zsh
pkviet: Tested with mbedtls 3.2.1; the datachannel dll worked fine.
There's no need to update mbedtls on windows account.

pkviet:
Regarding the libdatachannel commit that is checked out:
- I noticed an issue when using master on windows; there's a consistent 2 sec at start of a whip stream where frames are missed and dropped due to encoder lag.
- the issue was bisected and I picked the last good commit for windows, which is therefore used in that PR.
- the issue only occurs when building the libdatachannel dll with mingw and cross-compiling. I don't see the issue if libdatachannel is compiled on MSVC.
- the reason for the issue is unknown but seems related to threading( the 'bad' libdatachannel commits have to do with threading).

deps.ffmpeg/60-mbedtls.zsh
pkviet: clarification following internal discussion: in my tests, BOTH mbedtls 3.2.1 and 3.4.0 produced libdatachannel dlls which work with the current WHIP PR (using the C API of libdatachannel).

RytoEX:
> Regarding the libdatachannel commit that is checked out:
>     * I noticed an issue when using master on windows; there's a consistent 2 sec at start of a whip stream where frames are missed and dropped due to encoder lag.
>     * the issue was bisected and I picked the last good commit for windows, which is therefore used in that PR.
>     * the issue only occurs when building the libdatachannel dll with mingw and cross-compiling. I don't see the issue if libdatachannel is compiled on MSVC.
>     * the reason for the issue is unknown but seems related to threading( the 'bad' libdatachannel commits have to do with threading).

If there are specific commits in libdatachannel that are causing a regression, we can forcefully patch them out or revert them if needed in the patch step of the dependency script. It would be ideal if the issue could be fixed upstream though.

If I understand the conversation correctly, it is these changes to libjuice that are seemingly causing this issue of 2 seconds of blocking at the start of streams:
https://github.com/paullouisageneau/libdatachannel/commit/d3938a77697a55a285a26941c24a24e931f1046f
https://github.com/paullouisageneau/libjuice/compare/8c23cc88c6d41e5ccbc44ea0ad3d79b22cf02361...0d8c96650515d625139278d42d6ef4867dc48ce8

The regression is being tracked at https://github.com/paullouisageneau/libdatachannel/issues/892.

Is that about the sum of things? Has someone tried bisecting the libjuice changes to confirm which commit there causes the issue, or are we working off of a best guess?

Sean-Der:
@RytoEX Can we for now not take those future commits, since we don't need them? I will work with upstream and get them resolved. I don't have access to a Windows PC for ~2 weeks (when I return to work)

Talking with @pkviet I believe this PR is ready to be merged. If we can land the deps then we can work concurrently again. I will go work on upstream issue, while other devs address the obs-studio PR itself.

thank you

pkviet:
> If I understand the conversation correctly, it is these changes to libjuice that are seemingly causing this issue of 2 seconds of blocking at the start of streams:
> https://github.com/paullouisageneau/libdatachannel/commit/d3938a77697a55a285a26941c24a24e931f1046f
> https://github.com/paullouisageneau/libjuice/compare/8c23cc88c6d41e5ccbc44ea0ad3d79b22cf02361...0d8c96650515d625139278d42d6ef4867dc48ce8
> 
> The regression is being tracked at https://github.com/paullouisageneau/libdatachannel/issues/892.
> 
> Is that about the sum of things? Has someone tried bisecting the libjuice changes to confirm which commit there causes the issue, or are we working off of a best guess?

@RytoEX yes you understand correctly.
The regression appears here:
https://github.com/paullouisageneau/libjuice/commit/09ce20fee7cf637b46afd1450e3ff28fe65f09d0
It's unclear if it's a bug on libjuice side or a bug in winpthread since an msvc build doesn't display the 'bug'.

For now, leaving out the commits related to thread naming, doesn't pose any issues.
Note that i checked out to the last good commit, instead of reverting and i patched the lone commit merged after the 'thread name' commits.
For some reason i don't get, git revert -m 2 on the merge commit for the 'thread name' branch didn't work.

Some things to explore in the future:
- we could move the building of libdatachannel to msvc,
- we could try linking statically to libdatachannel.

For now, the dlls generated in this pr work fine along with the whip pr. Although i'd rather prefer the issue with 'thread names' fixed eventually, this PR is imo in a mergeable state.

deps.ffmpeg/60-mbedtls.zsh
RytoEX: This appears to have been fixed. @PatTheMav

deps.ffmpeg/70-libdatachannel.zsh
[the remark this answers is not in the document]
RytoEX: Is this still a concern?

utils.zsh/toolchain/windows-cross-toolchain.cmake
RytoEX: Looks like this has been addressed, @PatTheMav ?

Sean-Der:
@PatTheMav do my changes look good? Anything else I should be doing?

deps.ffmpeg/70-libdatachannel.zsh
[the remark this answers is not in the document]
PatTheMav: `BUILD_SHARED_LIBS` will only work if the library type hasn't been hardcoded in the `add_library` call.

PatTheMav:
I don't think my question regarding shared/static has been fully answered.

pkviet:
> I don't think my question regarding shared/static has been fully answered.
@PatTheMav 
Libdatachannel could be built as static since presumably it won't be used by another module.
I tried on windows + msvc to build mbedtls statically and then libdatachannel. I had issues though at the install step to create the right cmake target files.
Same issue if I cross compile on wsl + mingw.
Since the dll in the current pr works, i don't want to spend more time on investigating how to improve things with a fully static build, being neither the author of the PR nor that of the whip PR, although out of curiosity I did provide some help since the main authors can't easily develop or test on windows.
I would wait till the new FFmpeg msvc workflow is merged to seek for a more satisfactory solution. For the time being it's not ideal but works fine enough imo.

PatTheMav:
Sounds good to me then.

pkviet:
The libdatachannel situation would be the same as librist and SRT, which have the same mbedtls dependence and are shared builds.

deps.ffmpeg/60-mbedtls.zsh
```diff
@@ -53,7 +55,7 @@ patch() {
   for patch (${patches}) {
     read _target _url _hash <<< "${patch}"
 
-    if [[ ${_target} == "${target%%-*}" ]] apply_patch "${_url}" "${_hash}"
+    if [[ "${target%%-*}" == ${~_target} ]] apply_patch "${_url}" "${_hash}"
```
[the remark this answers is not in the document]
PatTheMav: The `${~<VAR>}` pattern turns the string contained in `<VAR>` into a glob expression. Glob expressions in turn need be on the right-hand side of a comparison.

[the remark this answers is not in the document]
RytoEX: Sounds good. Thanks for the explanation.

RytoEX (approved):
Feedback addressed. This seems good to go.

PR #174, opened 2023-03-29
deps.windows: Update dependencies for CMake 3.0 upgrade

PatTheMav:
### Description

* Disables x86 builds for all dependencies not required by x86-only modules.
* Moves luajit and Swig files into their canonical locations expected by default CMake find rules.
* Uses VSSetup PowerShell module to find Visual Studio installations in a more portable manner (also supports ARM64 Windows hosts)
* Creates obs-deps `VERSION` file in Windows packages

### Motivation and Context
* Enables compatibility with CMake 3.0 upgrade for Windows
* Removes complexity for CMake finders in said upgrade (canonical paths are used)
* Enables use of a single x64 project for Windows

### How Has This Been Tested?
Tested builds on Windows 11

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -375,21 +375,20 @@ jobs:
           outputName: qt6-${{ matrix.target }}-${{ needs.pre-checks.outputs.shortHash }}
 
   windows-build:
-    name: 'Build Windows Dependencies'
+    name: Build Windows Dependencies
     runs-on: windows-2022
+    needs: pre-checks
     strategy:
       fail-fast: true
       matrix:
         target: [x64, x86]
         include:
           - target: x64
-            config: 'Release'
-            type: 'static'
+            config: Release
+            type: static
           - target: x86
-            config: 'Release'
-            type: 'static'
-    env:
-      CACHE_REVISION: '3'
+            config: Release
+            type: static
     defaults:
       run:
         shell: pwsh
@@ -400,41 +399,26 @@ jobs:
       - name: Setup Environment
         id: setup
         run: |
+          # Setup Environment
+
           $Target='${{ matrix.target }}'
-          $ArtifactName="deps-windows-${Target}-${{ github.sha }}"
+          $ArtifactName="deps-windows-${Target}-${{ needs.pre-checks.outputs.shortHash }}"
           $FileName="windows-deps-$(Get-Date -Format 'yyyy-MM-dd')-${Target}.zip"
 
           "artifactName=${ArtifactName}" >> $env:GITHUB_OUTPUT
           "artifactFileName=${FileName}" >> $env:GITHUB_OUTPUT
 
-      - name: 'Check for GitHub Labels'
-        id: seekingTesters
-        if: ${{ github.event_name == 'pull_request' }}
-        run: |
-          $LabelFound = try {
-            $Params = @{
-              Authentication = 'Bearer'
-              Token = (ConvertTo-SecureString '${{ secrets.GITHUB_TOKEN }}' -AsPlainText)
-              Uri = '${{ github.event.pull_request.url }}'
-              UseBasicParsing = $true
-            }
-
-            (Invoke-RestMethod @Params).labels.name.contains("Seeking Testers")
-          } catch {
-            $false
-          }
-
-          "found=$(([string]${LabelFound}).ToLower())" >> $env:GITHUB_OUTPUT
-
-      - name: Build ntv2 debug
+      - name: Build ntv2 Debug
+        if: matrix.target == 'x64'
         shell: pwsh
         run: |
+          # Build ntv2 Debug
+
           $Params = @{
             Target = '${{ matrix.target }}'
             Configuration = 'Debug'
             Dependencies = 'ntv2'
           }
-          if ( '${{ matrix.type }}' -eq 'shared' ) { $Params += @{Shared = $true} }
 
           ./Build-Dependencies.ps1 @Params
           Remove-Item -Recurse -Force ${{ github.workspace }}/windows_build_temp
@@ -445,10 +429,9 @@ jobs:
           target: ${{ matrix.target }}
           type: ${{ matrix.type }}
           config: ${{ matrix.config }}
-          cacheRevision: ${{ env.CACHE_REVISION }}
 
       - name: Publish Build Artifacts
-        if: ${{ success() && (github.event_name != 'pull_request' || steps.seekingTesters.outputs.found == 'true') }}
+        if: github.event_name != 'pull_request' || fromJSON(needs.pre-checks.outputs.seekingTesters)
         uses: actions/upload-artifact@v3
         with:
           name: ${{ steps.setup.outputs.artifactName }}
--- a/Build-Dependencies.ps1
+++ b/Build-Dependencies.ps1
@@ -11,7 +11,8 @@ param(
     [switch] $SkipAll,
     [switch] $SkipBuild,
     [switch] $SkipDeps,
-    [switch] $SkipUnpack
+    [switch] $SkipUnpack,
+    [switch] $VSPrerelease
 )
 
 $ErrorActionPreference = "Stop"
--- a/Build-Dependencies.ps1
+++ b/Build-Dependencies.ps1
@@ -62,7 +62,15 @@ function Run-Stages {
 
         . $Dependency
 
-        if ( ! ( $Targets.contains($Target) ) ) { continue }
+        if ( ! ( $Targets.contains($Target) ) ) {
+            $Stages | ForEach-Object {
+                Log-Debug "Removing function $_"
+                Remove-Item -ErrorAction 'SilentlyContinue' function:$_
+                $script:StageName = ''
+            }
+
+            return
+        }
 
         if ( $Version -eq '' ) { $Version = $Versions[$Target] }
         if ( $Uri -eq '' ) { $Uri = $Uris[$Target] }
@@ -133,6 +141,19 @@ function Package-Dependencies {
     )
 
     Remove-Item -ErrorAction 'SilentlyContinue' -Path $Items -Force -Recurse
+
+    $Params = @{
+        ErrorAction = "SilentlyContinue"
+        Path = @(
+            "share/obs-deps"
+        )
+        ItemType = "Directory"
+        Force = $true
+    }
+
+    New-Item @Params *> $null
+
+    "$(Get-Date -Format "yyyy-MM-dd")" > share/obs-deps/VERSION
 
     Log-Information "Package dependencies"
 
--- a/deps.windows/50-swig.ps1
+++ b/deps.windows/50-swig.ps1
@@ -90,6 +90,14 @@ function Fixup {
     $Items = @(
         @{
             Path = "$($ConfigData.OutputPath)/swig/bin/swig.exe"
+            Destination = "$($ConfigData.OutputPath)/bin/swig.exe"
+        },
+        @{
+            Path = "$($ConfigData.OutputPath)/swig/share/swig/${Version}"
+            Destination = "$($ConfigData.OutputPath)/bin/Lib"
+        }
+        @{
+            Path = "$($ConfigData.OutputPath)/swig/bin/swig.exe"
             Destination = "$($ConfigData.OutputPath)/swig/swig.exe"
         },
         @{
@@ -101,7 +109,10 @@ function Fixup {
     $Items | ForEach-Object {
         $Item = $_
         Log-Status ('{0} => {1}' -f $Item.Path, $Item.Destination)
-        Move-Item @Item
+        if ( Test-Path $Item.Destination ) {
+            Remove-Item -Recurse -Force $Item.Destination
+        }
+        Copy-Item @Item -Recurse
     }
 
     Remove-Item "$($ConfigData.OutputPath)/swig/share" -Recurse -ErrorAction 'SilentlyContinue'
--- a/deps.windows/60-python.ps1
+++ b/deps.windows/60-python.ps1
@@ -10,30 +10,30 @@ param(
 )
 
 function Enable-PyEnv {
-    $Env:PYENV = "$(Get-Location | Convert-Path)\pyenv-win"
-    $Env:PYENV_ROOT = $Env:PYENV
-    $Env:PYENV_HOME = $Env:PYENV
+    $env:PYENV = "$(Get-Location | Convert-Path)\pyenv-win"
+    $env:PYENV_ROOT = $env:PYENV
+    $env:PYENV_HOME = $env:PYENV
 
-    $Env:OriginalPath = $Env:Path
+    $env:OriginalPath = $env:Path
 
-    $PathElements = ([Collections.Generic.HashSet[string]]::new([string[]]($Env:Path -split [System.IO.Path]::PathSeparator), [StringComparer]::OrdinalIgnoreCase))
+    $PathElements = ([Collections.Generic.HashSet[string]]::new([string[]]($env:Path -split [System.IO.Path]::PathSeparator), [StringComparer]::OrdinalIgnoreCase))
 
     $PyEnvPaths = @(
-        "${Env:PYENV}\bin"
-        "${Env:PYENV}\shims"
+        "${env:PYENV}\bin"
+        "${env:PYENV}\shims"
     )
-    $Env:Path = ($PyEnvPaths + $PathElements) -join [System.IO.Path]::PathSeparator
+    $env:Path = ($PyEnvPaths + $PathElements) -join [System.IO.Path]::PathSeparator
 
     $null = Invoke-External pyenv rehash
 }
 
 function Disable-PyEnv {
-    $Env:Path = $Env:OriginalPath
+    $env:Path = $Env:OriginalPath
 
-    Remove-Item Env:OriginalPath
-    Remove-Item Env:PYENV
-    Remove-Item Env:PYENV_ROOT
-    Remove-Item Env:PYENV_HOME
+    Remove-Item env:OriginalPath
+    Remove-Item env:PYENV
+    Remove-Item env:PYENV_ROOT
+    Remove-Item env:PYENV_HOME
 }
 
 function Setup {
@@ -105,6 +105,6 @@ function Install {
     $Items | ForEach-Object {
         $Item = $_
         Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
-        Copy-Item @Item
+        Copy-Item -Force @Item
     }
 }
--- a/utils.pwsh/Setup-Target.ps1
+++ b/utils.pwsh/Setup-Target.ps1
@@ -39,8 +39,8 @@ function Setup-BuildParameters {
     $VisualStudioData = Find-VisualStudio
 
     $VisualStudioId = "Visual Studio {0} {1}" -f @(
-        ([System.Version] $VisualStudioData.Version).Major
-        ([regex]::Match($VisualStudioData.Name, "\b\d{4}\b"))
+        $VisualStudioData.InstallationVersion.Major
+        ($VisualStudioData.DisplayName -split ' ')[-1]
     )
 
     $script:CmakeOptions = @(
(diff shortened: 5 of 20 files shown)


utils.pwsh/Setup-Target.ps1
```diff
@@ -77,29 +77,15 @@ function Find-VisualStudio {
...
-        }
-        $VisualStudioData = $VisualStudioDataTemp
+   $VSSetupData = Get-InstalledModule VSSetup -ErrorAction SilentlyContinue
+
+    if ( $VSSetupData -eq $null ) {
+        Install-Module VSSetup -Scope CurrentUser
     }
 
-    if ( ! ( $VisualStudioData ) -or ( $VisualStudioData.Version -lt 16 ) ) {
+    $VisualStudioData = Get-VSSetupInstance | Select-VSSetupInstance -Version '[16.0,18.0)' -Latest
```
RytoEX, line 86: If a user has VS2022 Stable and Preview both installed, which will it select? The system I would normally check this on is currently not up. The docs say, "By default, only releases are shown." unless you specify the `-Prerelease` flag, but a confirmation would be nice. If this does indeed only list stable releases, then this would be a slight behavior change, as we previously selected the first available from 2022 Release, 2022 Preview, 2019 Release. This should be fine on CI where only Release is available, though it could break on systems that only have VS2022 Preview installed.

utils.pwsh/Setup-Target.ps1
```diff
@@ -77,29 +77,15 @@ function Find-VisualStudio {
...
-            $VisualStudioDataTemp = ($VisualStudioData | Where-Object {$_.ChannelId -eq $SupportedVSVersions[$i]} | Sort-Object -Property Version)[0]
-            if ( $VisualStudioDataTemp ) {
-                break;
-            }
-        }
-        $VisualStudioData = $VisualStudioDataTemp
+   $VSSetupData = Get-InstalledModule VSSetup -ErrorAction SilentlyContinue
+
+    if ( $VSSetupData -eq $null ) {
+        Install-Module VSSetup -Scope CurrentUser
```
RytoEX, line 83: Are there any permissions required to install this module in this scope? Are there any known reasons it could fail to install? Should we handle a failure case, or will the script bail out on its own?

utils.pwsh/Setup-Target.ps1
```diff
@@ -39,8 +39,8 @@ function Setup-BuildParameters {
     $VisualStudioData = Find-VisualStudio
 
     $VisualStudioId = "Visual Studio {0} {1}" -f @(
-        ([System.Version] $VisualStudioData.Version).Major
-        ( $VisualStudioData.Name -split ' ')[3]
+        ([System.Version] $VisualStudioData.InstallationVersion).Major
```
RytoEX, line 42: Does this actually need the type specified, or is it just good practice? Checking both of these gives me the same output:
```
$VisualStudioData.InstallationVersion.GetType()
([System.Version] $VisualStudioData.InstallationVersion).GetType()
```

RytoEX (changes requested):

utils.pwsh/Setup-Target.ps1
PatTheMav: By default it should only pick up normal releases sorted by version and installation date. Prerelease versions are only included when the `-prerelease` or `-all` switch is added. This mirrors the behaviour of `vswhere` (which is on purpose iirc).

utils.pwsh/Setup-Target.ps1
PatTheMav: None listed by Microsoft - it might prompt for installation permission when run on a local system, but that'd be something I would leave to a user by design and not override.

utils.pwsh/Setup-Target.ps1
PatTheMav: It's good practise, because the cast makes it a version even if it's just a string - without it we'd have to check for the type before accessing the `Member` variable to avoid an exception.

RytoEX: I mean that it doesn't seem to materially change the type or output of this particular data returned from VSSetup, which is already a `Version`. By comparison, the type for `InstallationLocation` from the old method _was_ a string, so it was necessary there. I'm not super fussed about it, it just seemed unnecessary with the new method.
```
PS C:\> $VisualStudioData.InstallationVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
17     5      33530  505

PS C:\> $VisualStudioData.InstallationVersion.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Version                                  System.Object

> > ([System.Version] $VisualStudioData.InstallationVersion)

Major  Minor  Build  Revision
-----  -----  -----  --------
17     5      33530  505

PS C:\> ([System.Version] $VisualStudioData.InstallationVersion).GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Version                                  System.Object

PS C:\> $VisualStudioData = Get-CimInstance MSFT_VSInstance
PS C:\> $VisualStudioData[0].InstallLocation
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community

PS C:\> $VisualStudioData[0].InstallLocation.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     String                                   System.Object
```

utils.pwsh/Setup-Target.ps1
RytoEX: Sounds good to me.

utils.pwsh/Setup-Target.ps1
RytoEX: Hmm. Our old behavior checked for VS2022 Release, then VS2022 Preview, then VS2019 Release. If we want to mimic that behavior, we might have to call this multiple times for each specific version. Otherwise, we abandon that behavior and accept that it will fail on systems that only have VS2022 Preview installed.

PatTheMav: I think if we add `-prerelease` the choice would still be based on version number. So If VS2022 17.1 and VS2022 17.1 Preview are installed (not that this is a probable situation), I guess the release version is "more recent". VS2022 17.2 Preview would _be_ more recent however. So it would always chose whatever has the higher version number.

To that end we could just add that one to also include preview versions.

RytoEX: If 17.6 Release and 17.7 Preview were both installed and `-Prerelease -Latest` was specified, I imagine it would pick 17.7 Preview. The real question is what if the user has 17.7 Release and 17.7 Preview installed.

That said, the previous selection process didn't even bother taking this into account, it simply preferred:
1. VS2022 Release (any version)
2. VS2022 Preview (any version)
3. VS2019 Release (any version)

If you had VS2022 17.3 Release and VS2022 17.7 Preview installed, it would select VS23022 17.3 Release. This was, presumably, to allow devs to have the Release version that we targeted installed while also having some other Preview version installed if they needed it for other dev purposes outside of OBS.

Will think on it a bit.

utils.pwsh/Setup-Target.ps1
RytoEX: As an aside, this will probably cause a user prompt, which is fine, but something to be aware of.
```powershell
PS C:\> Install-Module VSSetup -Scope CurrentUser

Untrusted repository
You are installing the modules from an untrusted repository. If you trust this repository, change its
InstallationPolicy value by running the Set-PSRepository cmdlet. Are you sure you want to install the modules from
'PSGallery'?
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "N"):
```

utils.pwsh/Setup-Target.ps1
PatTheMav: Release and Preview would probably differ in their patch or build versions, so that would be taken into account. So if Release is build 6521 and Preview is 6532, then Preview "wins".

I would expect this to be a very simple version number comparison, only one variant includes the version numbers of preview builds and the other doesn't.

utils.pwsh/Setup-Target.ps1
PatTheMav: Correct. Personally I feel it's a good choice not to suppress this, but let users handle this on their local machines. It's in any case simpler than whatever script switches we could provide.

utils.pwsh/Setup-Target.ps1
RytoEX: Tested with 17.5 Release and 17.7 Preview.
```pwsh
PS C:\> Get-VSSetupInstance | Select-VSSetupInstance -Version '[16.0,18.0)' -Latest

InstanceId          : 142e8861
DisplayName         : Visual Studio Community 2022
InstallationVersion : 17.5.33627.172
InstallationPath    : C:\Program Files\Microsoft Visual Studio\2022\Community
InstallDate         : 11/1/2022 1:38:25 PM


PS C:\> Get-VSSetupInstance -Prerelease | Select-VSSetupInstance -Version '[16.0,18.0)' -Latest

InstanceId          : 3d2ad485
DisplayName         : Visual Studio Community 2022
InstallationVersion : 17.7.33711.374
InstallationPath    : C:\Program Files\Microsoft Visual Studio\2022\Preview
InstallDate         : 5/20/2023 1:40:40 PM
```

RytoEX: That at least worked as expected. It is still just a little unfortunate that it differs from current behavior. The only way I can think of to get the current behavior is to do multiple calls.

```pwsh
$VSInstance = Get-VSSetupInstance | Select-VSSetupInstance -Version '[16.0,18.0)' -Latest
// some null/empty/false check
$VSInstance = Get-VSSetupInstance -Prerelease | Select-VSSetupInstance -Version '[16.0,18.0)' -Latest
```

PatTheMav: I'd argue that the old behaviour was actually broken insofar as even a more recent VS 2022 preview version would not have been picked up.

The only scenario that code was made for was for the transitional period for when there was only a preview version of VS 2022 but a release was about to happen and we wanted the jobs not to suddenly break on CI.

RytoEX: > The only scenario that code was made for was for the transitional period for when there was only a preview version of VS 2022 but a release was about to happen and we wanted the jobs not to suddenly break on CI.

That's not what it was for. GitHub Actions images never have VS Preview builds, only Release builds. If I recall correctly, it was intended to provide build stability for _local_ builds where a dev might have multiple versions of VS installed for _some_ reason (work, other dev projects, etc.), but still wanted to be able to build obs-deps.

PatTheMav: Fixed via adding explicit prerelease switch to main build script toggling inclusion of Preview builds in version checks.

deps.windows/50-swig.ps1
```diff
@@ -90,20 +91,19 @@ function Fixup {
...
     $Items | ForEach-Object {
         $Item = $_
         Log-Status ('{0} => {1}' -f $Item.Path, $Item.Destination)
-        Move-Item @Item
+        Move-Item @Item -Force
     }
 
-    Remove-Item "$($ConfigData.OutputPath)/swig/share" -Recurse -ErrorAction 'SilentlyContinue'
-    Remove-Item "$($ConfigData.OutputPath)/swig/bin" -Recurse -ErrorAction 'SilentlyContinue'
+    Remove-Item "$($ConfigData.OutputPath)/swig" -Recurse -ErrorAction SilentlyContinue
```
RytoEX, line 108: Is there really nothing left in `$($ConfigData.OutputPath)/swig` to keep?

PatTheMav: Not that I'm aware of - anything swig needs is its binary and the language-specific modules which sit in the `share` subdirectory.

Afaik Swig also just looks for a `Lib` directory adjacent to the binary exclusively on Windows, because there is no POSIX directory standard - more recent versions might _also_ support that, but it was the search location with the highest priority.

RytoEX (approved):
Seems generally okay to me at a glance. Will test locally once my dev environment is working again. May merge a couple of other PRs first, but will try to keep it minimal.

utils.pwsh/Setup-Target.ps1
```diff
@@ -39,8 +39,8 @@ function Setup-BuildParameters {
     $VisualStudioData = Find-VisualStudio
 
     $VisualStudioId = "Visual Studio {0} {1}" -f @(
-        ([System.Version] $VisualStudioData.Version).Major
-        ([regex]::Match($VisualStudioData.Name, "\b\d{4}\b"))
+        $VisualStudioData.InstallationVersion.Major
+        ( $VisualStudioData.DisplayName -split ' ')[3]
```
RytoEX, line 43: This will need adjusted to not overwrite the change to use regex.

RytoEX (changes requested):

utils.pwsh/Setup-Target.ps1
PatTheMav: That change is intended - the old variant was based on the output of the CMI provider, the update uses `VSSetup` and its output instead.

@Absolucy would you be able to check what the `DisplayName` given by `VSSetup` is in your setup?

RytoEX: There's no need. I checked this already before merging #178.
```
PS C:\> Get-VSSetupInstance -Prerelease:$false

InstanceId          : 142e8861
DisplayName         : Visual Studio Community 2022
InstallationVersion : 17.6.33723.286
InstallationPath    : C:\Program Files\Microsoft Visual Studio\2022\Community
InstallDate         : 11/1/2022 1:38:25 PM

InstanceId          : a63227f3
DisplayName         : Visual Studio Build Tools 2022
InstallationVersion : 17.6.33723.286
InstallationPath    : C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools
InstallDate         : 5/30/2023 1:17:32 PM
```

deps.windows/60-python.ps1
```diff
@@ -10,30 +10,30 @@ param(
...
+    $PathElements = ([Collections.Generic.HashSet[string]]::new([string[]]($env:Path -split [System.IO.Path]::PathSeparator), [StringComparer]::OrdinalIgnoreCase))
 
     $PyEnvPaths = @(
-        "${Env:PYENV}\bin"
-        "${Env:PYENV}\shims"
+        "${env:PYENV}\bin"
+        "${env:PYENV}\shims"
     )
-    $Env:Path = ($PyEnvPaths + $PathElements) -join [System.IO.Path]::PathSeparator
+    $env:Path = ($PyEnvPaths + $PathElements) -join [System.IO.Path]::PathSeparator
```
RytoEX, line 25: Why the capitalization change here?

utils.pwsh/Setup-Target.ps1
```diff
@@ -77,29 +77,15 @@ function Find-VisualStudio {
...
-        "VisualStudio.16.Release"
-    $NumSupportedVSVersions = $SupportedVSVersions.length
-
-    if ( $VisualStudioData.GetType() -eq [object[]] ) {
-        for ( $i = 0; $i -lt $NumSupportedVSVersions; $i++ ) {
-            $VisualStudioDataTemp = ($VisualStudioData | Where-Object {$_.ChannelId -eq $SupportedVSVersions[$i]} | Sort-Object -Property Version)[0]
-            if ( $VisualStudioDataTemp ) {
-                break;
-            }
+    if ( $env:CI -eq $null ) {
```
RytoEX, line 80: `$Env` vs. `$env`?

utils.pwsh/Setup-Target.ps1
```diff
@@ -39,8 +39,8 @@ function Setup-BuildParameters {
     $VisualStudioData = Find-VisualStudio
 
     $VisualStudioId = "Visual Studio {0} {1}" -f @(
-        ([System.Version] $VisualStudioData.Version).Major
-        ([regex]::Match($VisualStudioData.Name, "\b\d{4}\b"))
+        $VisualStudioData.InstallationVersion.Major
+        ( $VisualStudioData.DisplayName -split ' ')[-1]
```
RytoEX, line 43: There should either be a space at the end of this parenthetical, or we should remove the space at the beginning. We're not 100% consistent with this throughout, but we don't need to do a consistency sweep in this PR.

deps.windows/60-python.ps1
```diff
@@ -10,30 +10,30 @@ param(
...
+    $env:Path = $Env:OriginalPath
 
-    Remove-Item Env:OriginalPath
-    Remove-Item Env:PYENV
-    Remove-Item Env:PYENV_ROOT
-    Remove-Item Env:PYENV_HOME
+    Remove-Item env:OriginalPath
+    Remove-Item env:PYENV
+    Remove-Item env:PYENV_ROOT
+    Remove-Item env:PYENV_HOME
```
RytoEX, line 36: What's the reason for this change?

Build-Dependencies.ps1
```diff
@@ -62,7 +63,15 @@ function Run-Stages {
...
-        if ( ! ( $Targets.contains($Target) ) ) { continue }
+        if ( ! ( $Targets.contains($Target) ) ) {
+            $Stages | ForEach-Object {
+                Log-Debug "Removing function $_"
+                Remove-Item -ErrorAction 'SilentlyContinue' function:$_
+                $script:StageName = ''
+            }
+
+            return
+        }
```
RytoEX, line 74: What is the intent here?

As an aside, although this `if` is not new, it seems like the `if` test will always return `! ( $true )`, since `$Targets` is set to `@($script:Target)` just above. Perhaps I've misunderstood something.

.github/workflows/main.yaml
```diff
@@ -400,41 +399,26 @@ jobs:
...
         shell: pwsh
         run: |
+          # Build ntv2 Debug
+
           $Params = @{
             Target = '${{ matrix.target }}'
             Configuration = 'Debug'
             Dependencies = 'ntv2'
           }
-          if ( '${{ matrix.type }}' -eq 'shared' ) { $Params += @{Shared = $true} }
```
RytoEX, line 437: What's the purpose of removing this?

RytoEX (changes requested):
A quick visual look over the files before I dig back into testing.

deps.windows/60-python.ps1
PatTheMav: Mainly that `env` (like `script` or other scopes) are not capitalised in Microsoft's documentation, and it also makes it aesthetically easier to read (as the _actually_ capitalised variable name stands out better).

utils.pwsh/Setup-Target.ps1
PatTheMav: See above.

utils.pwsh/Setup-Target.ps1
PatTheMav: I think we use spaces in conditionals, but don't use them when enclosing expressions.

deps.windows/60-python.ps1
PatTheMav: See above.

Build-Dependencies.ps1
PatTheMav: `continue` doesn't work as it would in a `foreach` blog - instead it continues the execution of the current script block and breaks the entire script.

`ForEach-Object` is literally borrowing behaviour from functional languages, in which the script block is an anonymous script block. Skipping an element thus requires simply returning from the current block early.

`$Stages` exists outside of that scope and is stateful, so an early return requires cleaning up all stage functions created by the dot import of the script file. Without this, the `Patch` function of an earlier dependency would be executed over and over again until a later dependency overwrites it with its own `Patch` step.

$Targets is overwritten with every dot import of a new dependency script and needs to be evaluated on every run of the anonymous script block.

.github/workflows/main.yaml
PatTheMav: The NTV2 SDK cannot be built as a shared library. Thus checking for the shared workflow variable is superfluous.

deps.windows/60-python.ps1
RytoEX: Discussed off-thread. Microsoft's documentation itself is inconsistent with capitalization.

While I'd prefer that style/consistency changes be done in a single sweep in their own PR, I don't want to hold this up with trying to excise and separate these changes.

Build-Dependencies.ps1
RytoEX: Discussed further off-thread.

RytoEX (approved):
Tested against obs-studio git and the OBS Studio Windows CMake 3 PR. Successfully built and ran OBS in both cases with my test scene collection, so this should be good to go.

PR #175, opened 2023-03-31
deps.qt: Backport macOS disabled combobox items fix

gxalpha:
### Description

Backports qt/qtbase@9f44553b3 which will only be in Qt 6.5.1 and later.

### Motivation and Context

Want disabled combobox items to actually be disabled.
Fixed it in Qt dev branch and cherry-picked to 6.5, but we're on 6.4 so it won't be there until we update again, so backport until then.

### How Has This Been Tested?

Tested with an earlier version of this patch.
CI builds would be required for a test with this exact version (and built from obs-deps).

### Types of changes

### Checklist:


PatTheMav (approved):

RytoEX:
Does the obs-studio PR mean that even if we backport this, obs-studio still performs the workaround rather than the fixed codepath? If so, what is the benefit to backporting this?

gxalpha:
The obs-studio workaround is only to check for when the unselectable item has already been selected (which with the backport cannot happen anymore). The code path would still be hit, but doesn't do anything (and doesn't hurt either).

RytoEX (approved):

RytoEX:
Reminder to ourselves that we can drop this patch when we start building 6.5.1 or later.

PR #178, opened 2023-04-04
utils.pwsh: Fix CMake error with VS Build Tools

Absolucy:
### Description
This prevents an error where the Visual Studio generator extraction thing would work incorrectly if you had only the Build Tools installed.

### Motivation and Context
If you have the Visual Studio Build Tools installed, and not a full copy of Visual Studio, building cmake things would fail, as the output of `Find-VisualStudio` would be something along the lines of this:
```
Caption           : Visual Studio Build Tools 2022
Description       : The Visual Studio Build Tools allows you to build native and managed MSBuild-based applications
                    without requiring the Visual Studio IDE. There are options to install the Visual C++ compilers and
                    libraries, MFC, ATL, and C++/CLI support.
ElementName       : Visual Studio Build Tools 2022
InstanceID        : Microsoft:VisualStudio:e66d2ae7
IdentifyingNumber : e66d2ae7
Name              : Visual Studio Build Tools 2022
SKUNumber         :
Vendor            : Microsoft Corporation
Version           : 17.5.33516.290
WarrantyDuration  :
WarrantyStartDate :
ChannelId         : VisualStudio.17.Release
ChannelUri        : https://aka.ms/vs/17/release/channel
InstallDate       : 3/26/2023 2:20:23 PM
InstallLocation   : C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools
IsComplete        : True
IsLaunchable      : True
IsPrerelease      : False
LayoutPath        :
ProductId         : Microsoft.VisualStudio.Product.BuildTools
ProductLocation   : C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\Tools\LaunchDevCmd.bat
State             : 4294967295
PSComputerName    :
```

Which means `( $VisualStudioData.Name -split ' ')[3]` would return `Tools`, resulting in the build script trying to use "Visual Studio 17 Tools" as a cmake generator, which is invalid.

I replaced it with a small regex that just extracts a 4-digit number (a year) from the name instead.

### How Has This Been Tested?
By running `Build-Dependencies.ps1` locally

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Setup-Target.ps1
+++ b/utils.pwsh/Setup-Target.ps1
@@ -40,7 +40,7 @@ function Setup-BuildParameters {
 
     $VisualStudioId = "Visual Studio {0} {1}" -f @(
         ([System.Version] $VisualStudioData.Version).Major
-        ( $VisualStudioData.Name -split ' ')[3]
+        ([regex]::Match($VisualStudioData.Name, "\b\d{4}\b"))
     )
 
     $script:CmakeOptions = @(


RytoEX (approved):
Have tested this locally against normal VS installations, and tested it on a system with both VS Build Tools and normal VS installations. I haven't seen another way to handle this, either in the current method of selecting a VS instance or in the reworked method in #174.

PR #182, opened 2023-04-21
Add Libqrcodegencpp

tytan652:
### Description

Some Linux distribution (e.g. Ubuntu, Fedora) tweak https://github.com/nayuki/QR-Code-generator to ship it's C and C++ implementations as libraries.

One of the methods is to add files from https://github.com/EasyCoding/qrcodegen-cmake and use CMake to build the libraries which is the method took for this PR.

The C implementation (libqrcodegen) is removed since it is not needed.

I'm aware of #174, so if it get merged first, I will update the pwsh script to add `$Targets`.

### Motivation and Context

Reduce the number of submodules in obs-studio repo and provides those as libraries in obs-deps.
Also this would allow to make obs-websocket submodule-less.

### How Has This Been Tested?

CI test build static version, and tested shared on VMs.

https://github.com/tytan652/obs-studio/pull/13 provides binaries:
 - Ubuntu 20.04 is failed too because of an API break from Libqrcodegen between 1.5.0 (20.04) and 1.7.0 (22.04) (header file name change), so one more reason to drop Ubuntu 20.04.

### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/60-qrcodegencpp.ps1
+++ b/deps.windows/60-qrcodegencpp.ps1
@@ -1,0 +1,111 @@
+param(
+    [string] $Name = 'qrcodegencpp',
+    [string] $Version = '1.8.0',
+    [string] $Uri = 'https://github.com/nayuki/QR-Code-generator.git',
+    [string] $Hash = '720f62bddb7226106071d4728c292cb1df519ceb',
+    [string] $UriCMake = 'https://github.com/EasyCoding/qrcodegen-cmake.git',
+    [string] $HashCMake = 'c57623d48a2d422b0f908dcf22d56d603e56c0e4'
+)
+
+function Setup {
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath "$Path/src"
+    Setup-Dependency -Uri $UriCMake -Hash $HashCMake -DestinationPath "$Path/qrcodegen-cmake"
+}
+
+function Clean {
+    Set-Location "$Path/src"
+
+    if ( Test-Path "build_${Target}" ) {
+        Log-Information "Clean build directory (${Target})"
+        Remove-Item -Path "build_${Target}" -Recurse -Force
+    }
+}
+
+function Patch {
+    Log-Information "Patch (${Target})"
+    Set-Location $Path
+
+    $Items = @(
+        @{
+            Path = "qrcodegen-cmake/CMakeLists.txt"
+            Destination = "src"
+            ErrorAction = 'SilentlyContinue'    
+        }
+        @{
+            Path = "qrcodegen-cmake/cmake"
+            Destination = "src"
+            Recurse = $true
+            ErrorAction = 'SilentlyContinue'    
+        }
+    )
+
+    $Items | ForEach-Object {
+        $Item = $_
+        Copy-Item @Item
+    }
+}
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location "$Path/src"
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
+    )
+
+    Invoke-External cmake -S . -B "build_${Target}" @Options
+}
+
+function Build {
+    Log-Information "Build (${Target})"
+    Set-Location "$Path/src"
+
+    $Options = @(
+        '--build', "build_${Target}"
+        '--config', $Configuration
+    )
+
+    if ( $VerbosePreference -eq 'Continue' ) {
+        $Options += '--verbose'
+    }
+
+    Invoke-External cmake @Options
+}
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location "$Path/src"
+
+    $Options = @(
+        '--install', "build_${Target}"
+        '--config', $Configuration
+    )
+
+    if ( $Configuration -match "(Release|MinSizeRel)" ) {
+        $Options += '--strip'
+    }
+
+    Invoke-External cmake @Options
+}
+
+function Fixup {
+    Log-Information "Fixup (${Target})"
+    Set-Location $Path
+
+    $Items = @(
+        @{ FullPath = "$($ConfigData.OutputPath)/include/qrcodegen" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/cmake/qrcodegen" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/pkgconfig/qrcodegen.pc" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/qrcodegen.*" }
+        @{ FullPath = "$($ConfigData.OutputPath)/bin/qrcodegen.*" }
+    )
+
+    $Items | ForEach-Object {
+        $Item = $_
+        if ( Test-Path $Item.FullPath ) {
+            Remove-Item -Recurse -Force $Item.FullPath
+        }
+    }
+}
--- a/deps.macos/80-qrcodegencpp.zsh
+++ b/deps.macos/80-qrcodegencpp.zsh
@@ -1,0 +1,109 @@
+autoload -Uz log_debug log_error log_info log_status log_output
+
+## Dependency Information
+local name='qrcodegencpp'
+local version='1.8.0'
+local url='https://github.com/nayuki/QR-Code-generator.git'
+local hash='720f62bddb7226106071d4728c292cb1df519ceb'
+local url_cmake='https://github.com/EasyCoding/qrcodegen-cmake.git'
+local hash_cmake='c57623d48a2d422b0f908dcf22d56d603e56c0e4'
+
+## Build Steps
+setup() {
+  log_info "Setup (%F{3}${target}%f)"
+  setup_dep ${url} ${hash}
+
+  mkdir -p qrcodegen-cmake-${version}
+  pushd qrcodegen-cmake-${version}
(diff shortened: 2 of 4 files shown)


Conan-Kudo:
I have a patch downstream in Fedora that [adapts the websocket plugin to use the system version of qrcodegencpp](https://src.fedoraproject.org/rpms/obs-studio/blob/rawhide/f/obs-studio-websocket-use-system-qrcodegencpp.patch). Once this is merged, would it be acceptable to get this upstreamed too?

tytan652:
> I have a patch downstream in Fedora that [adapts the websocket plugin to use the system version of qrcodegencpp](https://src.fedoraproject.org/rpms/obs-studio/blob/rawhide/f/obs-studio-websocket-use-system-qrcodegencpp.patch). Once this is merged, would it be acceptable to get this upstreamed too?

Read until the end.

No, because your patch rely on installed cmake modules (made by qrcodegen-cmake) which is:
1. non-existent in Ubuntu packages
2. has a bug with a non-working namespace (`qrcodegencpp::qrcodegencpp` does not work on my testing)

I created a new finder that works for any cases that use qrcodegen.hpp. https://github.com/PatTheMav/cmake-finders/pull/2 which will be added to OBS Studio later.

And here is the branch for my changes for obs-websocket: https://github.com/tytan652/obs-websocket/tree/last_ws_dep_standing

I discovered the existence of libqrcodegencpp package through you patch and thank you for that. But it does not work in all cases, so I had create something that will work almost anywhere.

Conan-Kudo:
@xvitaly, could you take a look at the bug @tytan652 discovered about your CMake overlay?

tytan652:
The bug has magically disappeared because of a reboot apparently, CMake why ?.
Sorry @xvitaly for the false alarm.

tytan652:
Sorry again, @xvitaly I can reproduce it, we are in a transition period in our CMake build-system and I modified the wrong file for Linux. I will write a issue on the qrcodegen-cmake repo rather than continuing the off-topic here.

xvitaly:
> I will write a issue on the qrcodegen-cmake repo rather than continuing the off-topic here.

Great. I will look into it.

deps.windows/60-qrcodegencpp.ps1
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    Invoke-External cmake @Options
+}
+
+function Fixup {
+    Log-Information "Fixup (${Target})"
+    Set-Location $Path
+
+    Remove-Item -ErrorAction 'SilentlyContinue' -Recurse "$($ConfigData.OutputPath)/include/qrcodegen"
```
PatTheMav, line 97: Either use `Get-ChildItems` to gather the files for deletion and pipe its result to `Remove-Item` once, or generate a hash table with all the deletion jobs and pipe it to `Remove-Item`:

```
Get-ChildItem ./lib/* -Include 'qrcodegen.*' | Remove-Item -Force -Recurse

[or]

$Items | ForEach-Object {
    $Item = $_
    if ( Test-Path $Item.FullPath ) {
        Remove-Item -Recurse -Force $Item.FullPath
    }
}
```

tytan652: Tried to do the hash table.

deps.windows/60-qrcodegencpp.ps1
```diff
@@ -0,0 +1,111 @@
...
+    $Items = @(
+        @{ FullPath = "$($ConfigData.OutputPath)/include/qrcodegen" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/cmake/qrcodegen" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/pkgconfig/qrcodegen.pc" }
+        @{ FullPath = "$($ConfigData.OutputPath)/lib/qrcodegen.*" }
+        @{ FullPath = "$($ConfigData.OutputPath)/bin/qrcodegen.*" }
+    )
+
+    $Items | ForEach-Object {
+    $Item = $_
```
PatTheMav, line 106: Needs indentation.

tytan652: Sorry, I only see this review weeks later.

Fixed.

PatTheMav (approved):

tytan652:
Commits:
- deps.macos: Add Libqrcodegencpp to macOS deps
- [TO SQUASH] Apply recent macOS deps changes
have been squashed since changes were approved.

RytoEX:
I almost wish that instead of being in a single script, this were two separate scripts, like:
```
deps.macos/80-qrcodegencpp.zsh
deps.macos/81-qrcodegencmake.zsh
deps.windows/60-qrcodegencpp.ps1
deps.windows/61-qrcodegencmake.ps1
```
But I'm not sure that would necessarily be any "better" than what is already here, and it might be more convoluted to get that working, so this is fine as-is if @PatTheMav is already fine with it.

tytan652:
> I almost wish that instead of being in a single script

`qrcodegen-cmake` is a big "patch" meant to modify the the `Qr-Code-Generator` to build as a standalone lib with CMake.
They can't be separated since we can't build `Qr-Code-Generator` without `qrcodegen-cmake`.

PatTheMav:
That's actually an oversight on my end - the CMake files need to be added via a patch file, not via a second checkout (and using relative path patterns like `mkcd ../something` is also no bueno, instead `pushd`/`popd` pairs should be used, but when using a proper patch file to generate the required CMake files directly in the source dir this should be unnecessary anyway).

PatTheMav:
> I will try to avoid the "relative path", but I will not increase the maintenance cost by generating a patch from a git repo meant to be copied inside another repo.

That's the canonical way we add CMake support to projects that don't have it. The other option would be a fork that has that patch applied. It seems very roundabout to me to have one repository with the sources, and another repository with the build system files, especially as you cannot combine both into the same directory.

RytoEX:
> > I will try to avoid the "relative path", but I will not increase the maintenance cost by generating a patch from a git repo meant to be copied inside another repo.
> 
> That's the canonical way we add CMake support to projects that don't have it. The other option would be a fork that has that patch applied. It seems very roundabout to me to have one repository with the sources, and another repository with the build system files, especially as you cannot combine both into the same directory.

Unfortunately, the CMake repo itself recommends this method:
```
git clone https://github.com/nayuki/QR-Code-generator.git qrcodegen
git clone https://github.com/EasyCoding/qrcodegen-cmake.git qrcodegen-cmake
ln -s ../qrcodegen-cmake/{cmake,CMakeLists.txt} qrcodegen/
mkdir -p qrcodegen-build
cmake -S qrcodegen -B qrcodegen-build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_EXAMPLES:BOOL=ON -DBUILD_TESTS:BOOL=ON
cmake --build qrcodegen-build
```
So as far as that project is concerned, git cloning both projects is correct.

PatTheMav:
> > > I will try to avoid the "relative path", but I will not increase the maintenance cost by generating a patch from a git repo meant to be copied inside another repo.
> > 
> > 
> > That's the canonical way we add CMake support to projects that don't have it. The other option would be a fork that has that patch applied. It seems very roundabout to me to have one repository with the sources, and another repository with the build system files, especially as you cannot combine both into the same directory.
> 
> Unfortunately, the CMake repo itself recommends this method:
> 
> ```
> git clone https://github.com/nayuki/QR-Code-generator.git qrcodegen
> git clone https://github.com/EasyCoding/qrcodegen-cmake.git qrcodegen-cmake
> ln -s ../qrcodegen-cmake/{cmake,CMakeLists.txt} qrcodegen/
> mkdir -p qrcodegen-build
> cmake -S qrcodegen -B qrcodegen-build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS:BOOL=ON -DBUILD_EXAMPLES:BOOL=ON -DBUILD_TESTS:BOOL=ON
> cmake --build qrcodegen-build
> ```
> 
> So as far as that project is concerned, git cloning both projects is correct.

Yeah "one within the other" is not possible (except for downloading the tarball from GitHub and extracting it in place). I don't like it, but the implementation in the PR is fine.

RytoEX:
> Yeah "one within the other" is not possible (except for downloading the tarball from GitHub and extracting it in place). I don't like it, but the implementation in the PR is fine.

Right. I think we can explore an alternative implementation in the future, or encourage the base project to add proper CMake support.

xvitaly:
> The other option would be a fork that has that patch applied. It seems very roundabout to me to have one repository with the sources, and another repository with the build system files, especially as you cannot combine both into the same directory.

The QR-Code-generator upstream is too unfriendly to external contributors. They don't accept pull requests. Our PR with full CMake support: https://github.com/nayuki/QR-Code-generator/pull/140

That's why we created our own repo with CMake support.

RytoEX (changes requested):
Whitespace nit.

deps.windows/60-qrcodegencpp.ps1
```diff
@@ -0,0 +1,111 @@
...
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location "$Path/src"
+
+	$OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
+    )
```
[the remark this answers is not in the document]
tytan652: Yes, a tab got here. Fixed.

RytoEX (approved):
Seems fine. Just need to determine merge order with respect to other pending PRs.

PR #183, opened 2023-05-04
Add creation of macOS dSYM debug symbols to symbolicate release builds

PatTheMav:
### Description
Updates all macOS-related build scripts to generate dSYM files in Release configuration.

Also contains a code-cleanup of the zsh-based scripts:

* Unnecessary quoting removed (zsh does not automatically split strings into fields, making space-separated paths safer to use)
* Due to the need of dSYM creation binary stripping is now explicitly done by each `fixup` step
* Shared library artefacts are automatically removed if a library is rebuilt as a static library (avoids leakage of prior shared library builds)
* C++ and C standard raised to C++17 and C17 respectively, except where the project is not compatible with the standard
* Cleaned up some code doing the same thing but having been implemented in separate ways (e.g. the way static library builds were enforced)
* Refreshed the clean-up steps in the main build scripts to remove more static libraries when a "complete" package is built (necessary because the static libraries are not stripped and thus increased in size)

### Motivation and Context
As obs-deps and obs-studio use Release configuration to ship the application, it is necessary to also ship the associated dSYM debug symbols used to build the application.

Debug symbols are identified by UUIDs that are linked to UUIDs of the built binaries, so only a matching pair is able to symbolicate crashes.

### How Has This Been Tested?
Debug symbols created for macOS-deps, FFmpeg, and Qt6, universal debug symbols created locally as well.

### Types of changes

### Checklist:


RytoEX:
> Updates all macOS-related build scripts to generate dSYM files in Release configuration.
> 
> Also contains a code-cleanup of the zsh-based scripts:

I kind of wish these were separate PRs so that it was easier to review the "code-cleanup" portions from the more significant changes, but I don't know how difficult it would be to separate them at this point.

PatTheMav:
> > Updates all macOS-related build scripts to generate dSYM files in Release configuration.
> > Also contains a code-cleanup of the zsh-based scripts:
> 
> I kind of wish these were separate PRs so that it was easier to review the "code-cleanup" portions from the more significant changes, but I don't know how difficult it would be to separate them at this point.

Most (but not all) are interconnected, as I had to touch all files anyway to add the dSYM creation step, which required adjusting the install steps for all CMake-based dependencies, etc.

PatTheMav:
Updated this and the other PRs to follow a single dependency trail:

1. https://github.com/obsproject/obs-deps/pull/188
2. https://github.com/obsproject/obs-deps/pull/174
3. https://github.com/obsproject/obs-deps/pull/183
4. https://github.com/obsproject/obs-deps/pull/186

Following this order will update each part of `obs-deps` one by one, such as that almost all parts of it are updated in some way at the end of the process.

deps.macos/80-pcre2.zsh
```diff
@@ -90,13 +81,29 @@ install() {
...
   )
 
-  if [[ "${config}" =~ "Release|MinSizeRel" ]] args+=(--strip)
   if (( _loglevel > 1 )) args+=(--verbose)
 
-  cd "${dir}"
+  cd ${dir}
   progress cmake ${args}
 }
+
```
RytoEX, line 93: Extra blank line? Should probably check other places where `fixup()` was added.

.github/actions/build-deps/action.yml
```diff
@@ -42,12 +49,30 @@ runs:
           !${{ inputs.workingDirectory }}/*_build_temp/*.tar.gz
           !${{ inputs.workingDirectory }}/*_build_temp/*.tar.xz
           !${{ inputs.workingDirectory }}/*_build_temp/*.zip
+          !${{ inputs.workingDirectory }}/*_build_temp/qt*
```
RytoEX, line 52: Why is Qt being added here?

.github/actions/build-deps/action.yml
```diff
@@ -69,4 +94,5 @@ runs:
...
```
RytoEX, line 97: Same question about Qt.

deps.ffmpeg/20-libpng.zsh
```diff
@@ -101,27 +101,58 @@ install() {
...
-      create_importlibs ${target_config[output_dir]}/bin/libpng*.dll(:a)
-    }
+  log_info "Fixup (%F{3}${target}%f)"
 
-    rm ${target_config[output_dir]}/bin/(libpng*-config|png*fix*)(N)
+  local strip_tool
+  local -a strip_files
+
+  case ${target} {
+    macos*)
```
RytoEX, line 123: `macos*` vs. `macos-*`?

deps.ffmpeg/50-libtheora.zsh
```diff
@@ -121,43 +121,66 @@ build() {
...
-        ;;
-      windows-x*)
+  cd ${dir}
+
+  log_info "Fixup (%F{3}${target}%f)"
+
+  local strip_tool
+  local -a strip_files
+
+
```
RytoEX, line 148: Extra blank line.

deps.ffmpeg/60-mbedtls.zsh
```diff
@@ -73,22 +73,24 @@ config() {
     -DGEN_FILES=OFF
   )
 
+  if [[ ${config} == Release ]] args+=(-DCMAKE_C_FLAGS="${c_flags} -std=c17 -g")
```
RytoEX, line 76: Do we only set c17 in Release?

RytoEX (changes requested):

.github/actions/build-deps/action.yml
PatTheMav: Probably a remnant of an older build environment on CI.

deps.ffmpeg/60-mbedtls.zsh
PatTheMav: No, but the "global" `CMAKE_C_FLAGS` variable is defined as `"${c_flags} -std=c17"` and if we want to append to that (as we need to do for Release to ensure dSYMs are generated), we need to replace the entire value on the command line (and thus repeat it here).

.github/workflows/main.yaml
```diff
@@ -161,6 +165,13 @@ jobs:
           name: ${{ steps.setup.outputs.artifactName }}
           path: ${{ github.workspace }}/${{ matrix.target }}/${{ steps.setup.outputs.artifactFileName }}
 
+      - name: Publish Debug Symbol Artifacts
+        if: ${{ startsWith(matrix.os, 'macos') && success() && (github.event_name != 'pull_request' || steps.seekingTesters.outputs.found == 'true') }}
```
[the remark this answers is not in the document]
RytoEX: If that's the case, is there a reason we're still checking it here? I think there are some weird edge cases where checking `success()` can produce different outcomes because the implicit `success()` call returns a different result than the explicit `success()` call, but I'm drawing a blank on the exact scenarios right now.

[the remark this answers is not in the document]
PatTheMav: Oh right, that's an oversight then.

deps.ffmpeg/20-libpng.zsh
RytoEX: Okay then.

deps.ffmpeg/60-mbedtls.zsh
PatTheMav: `CMAKE_C_FLAGS` are usually empty, instead the configuration-specific flags are set.

CMake itself adds `-g` (which enables debug information) for Debug and RelWithDebInfo only.

On macOS it is canonical to _always_ have debug symbols, but embed them in Debug and RelWithDebInfo (`DWARF`) and compile them into separate debug symbols (`DWARF with dSYM`) otherwise. Alas CMake does not support this behaviour.

Instead the `-g` flag needs to be added to compilation flags (as CMake itself checks for its presence in the flags to enable debug symbol generation). This happens inside CMake's generators, so generator-specific options are ignored.

In `30-build.zsh` we do just that: `-DCMAKE_C_FLAGS_RELEASE="-O3 -g -DNDEBUG"`.

MbedTLS however sets/overwrites `CMAKE_C_FLAGS_RELEASE`, so we lose this information. And the only way to get the `-g` flag back is to add it to the "global" `CMAKE_C_FLAGS` (which are empty in MbedTLS' case) for the release config only.

.github/workflows/main.yaml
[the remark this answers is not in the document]
RytoEX:
```
if: ${{ startsWith(matrix.os, 'macos') && success() && (github.event_name != 'pull_request' || steps.seekingTesters.outputs.found == 'true') }}
```
vs.
```
if: ${{ success() && startsWith(matrix.os, 'macos') && (github.event_name != 'pull_request' || steps.seekingTesters.outputs.found == 'true') }}
```

My other comment about it not being necessary is that it's applied by default unless you call another status check function. See: https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions:
> A default status check of success() is applied unless you include one of these functions.

If, for some reason, we want to highlight that this step only runs on `success` status, we can leave this here. In that case, I'm used to seeing it as the first conditional, hence my original comment.

This is a newly added line, so I'm not sure what this means:
> this PR will only add the necessary bits to get the dSYM files packaged, but leaves the rest as-is.

deps.ffmpeg/60-mbedtls.zsh
RytoEX: I think I understand.

.github/workflows/main.yaml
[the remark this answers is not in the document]
PatTheMav: As discussed off-thread, the intention of the change was to keep the format used overall in the FFmpeg build job, as the job itself will see a full update in https://github.com/obsproject/obs-deps/pull/186.

RytoEX (approved):

PR #184, opened 2023-05-22
Add HTTP/2 and Brotli support

derrod:
### Description

Adds brotli and nghttp2 libraries to curl for HTTP/2 and brotli compression support. Also switches curl to being a static library.

### Motivation and Context

Currently all HTTP traffic OBS receives is uncompressed, which is wasting bandwidth. Additionally, we want to use HTTP/2 for multiplexing and asnychronous requests in the updater and possibly other parts of OBS.

### How Has This Been Tested?

Built locally, but needs to be checked more thouroughly as updater changes are still in progress and we might need to build curl with a static VCRT for it (and thus probably twice in total), so the PR is a draft.

### Types of changes

### Checklist:


RytoEX:
> > Also switches curl to being a static library.
> 
> Unfortunately, OBS Studio is not the only binary relying on curl, some plugins do. So building statically is not a good thing, it will introduce duplication of the curl lib in various binaries.

Please be specific. What plugins rely on curl?

tytan652:
> What plugins rely on curl?

`windows-capture` and `rtmp-services`, they use it through the `file-updater`.

derrod:
> Unfortunately, OBS Studio is not the only binary relying on curl, some plugins do. So building statically is not a good thing, it will introduce duplication of the curl lib in various binaries.

Honestly, not worth caring about. It's a few kilobytes or maybe a megabyte. But it's either that or compiling Curl twice.

derrod:
Just to leave some notes on where this PR is at:
- libcurl's cmake is annoying because it exports with absolute paths to dependent libraries (nghttp2/brotli)
    + Probably requires some patch to their cmake to fix
    + We'd probably have to write finders as well for use in OBS
- Does compile and HTTP/2 works in OBS with minimal changes if you manually fix linking
- static+shared is still annoying, libcurl itself would probably support both as separate targets but its dependencies might not, haven't investigated

RytoEX:
> Just to leave some notes on where this PR is at:
>     * libcurl's cmake is annoying because it exports with absolute paths to dependent libraries (nghttp2/brotli)
>       * Probably requires some patch to their cmake to fix
>       * We'd probably have to write finders as well for use in OBS
>     * Does compile and HTTP/2 works in OBS with minimal changes if you manually fix linking
>     * static+shared is still annoying, libcurl itself would probably support both as separate targets but its dependencies might not, haven't investigated

On libcurl, I would prefer that if it needs changes, that those changes are done on the curl repo. They release pretty frequently these days.

derrod:
Based on some discussions off-thread we can probably use the dynamic CRT instead of static even for the updater, which makes things a little easier (see my [standalone updater repo](https://github.com/derrod/obs-updater)).

I don't even know where to start with the cmake changes or if upstream would even care about them.

PR #186, opened 2023-05-24
deps.ffmpeg: Switch Windows builds to native build toolchain

PatTheMav:
### Description
Adds an entirely Windows-based toolchain for building FFmpeg and its dependencies and replaces the current Linux-based toolchain.

### Motivation and Context
Adds the `ffmpeg` package to Windows build scripts (selected by providing `-PackageName ffmpeg` to the build script).

While building Windows binaries via MINGW64, it leads to several small issues:

* MINGW64 does not support C++11 threading natively, instead it is implemented via POSIX threads, which makes all code depend on them (and in our case requires linking POSIX threads statically)
* Some 3rd party projects only support POSIX threads when they are not built via an MSVC-based toolchain, so once again a non-native threading implementation is uses
* Debug information in program database format (`PDB`) is not supported by the toolchain entirely as `clang`/`gcc` always embed debug information and there is no MING64 variant of the program database server (the service that collects symbols during compilation and generates `PDB`s)
* It creates another abstraction between the compiled code and its target platform - MINGW64 will always lag behind the actual feature level of the Windows platform

This PR builds all(!) current FFmpeg dependencies and FFmpeg itself natively on Windows with the MSVC toolchain (using either `cl` or `clang-cl` as compiler):

* `zlib`: Built natively
* `libpng`: Built natively
* `opus`: Built natively
* `ogg`: Built natively
* `vorbis`: Built natively
* `libvpx`: Uses shell support and `make` from `MSYS2`, compiled with `MSBuild`
* `SVT-AV1`: Built natively
* `aom`: Built natively
* `x264`: Uses shell support and `make` from `MSYS2`, compiled with `cl`
* `libtheora`: Built natively
* `lame`: Built natively using old (but still working) `nmake` script
* `mbedtls`: Built natively, but due to its code architecture as static libraries only
* `srt`: Built natively, using C++11 threading and static `mbedtls`
* `librist`: Built natively, using `meson` in a VS DevShell and embedded `mbedtls`
* `ffmpeg`: Built natively with Win32 threading support, statically linking all 3rd party libraries except for `srt`, `librist`, and `x264` which are available as shared libraries

**Note**: The most impacting change on `obs-deps` is the need to also link to `mbedtls` statically due to the reasons above.

The scripts are entirely PowerShell-based, build dependencies are installed via `winget`, except for `meson` and `ninja` which are installed via `pip3` on GitHub Action runners.

POSIX tools (`pkg-config`, `yasm`, `nasm`, `make`, and others) are installed via `pacman` inside the `MSYS2` environment automatically.

### How Has This Been Tested?
Tested on CI and locally on Windows 11.

### Types of changes

### Checklist:


PatTheMav:
Note: This PR does _not_ add PDB creation - while some packages will automatically create PDBs because they are CMake-based (and thus use the MSVC toolchain natively), some do not (namely X264, libvpx, and FFmpeg itself), and others are built using clang-cl which does not generate PDBs (AOM).

It does however create a better starting point for adding Windows PDBs (see also https://github.com/obsproject/obs-deps/pull/183 which does that for macOS).

PatTheMav:
Updated this and the other PRs to follow a single dependency trail:

1. https://github.com/obsproject/obs-deps/pull/188
2. https://github.com/obsproject/obs-deps/pull/174
3. https://github.com/obsproject/obs-deps/pull/183
4. https://github.com/obsproject/obs-deps/pull/186

Following this order will update each part of `obs-deps` one by one, such as that almost all parts of it are updated in some way at the end of the process.

qcvndemo01:
Hi, I'm try testing this pull request but failed to build. Please help to guide me. Here is the detailed info:

1. Environment
- ARM64-based device
- Windows 11 Pro 22H2
- MSYS2, zsh, and ohmyzsh installed
- MSVC toolchain (visual studio 2022, etc.)
2. build
- clone the source code: _git clone https://github.com/PatTheMav/obs-deps.git -b ffmpeg-on-windows_
- Open MSYS2 MSYS, which run zsh shell by default
- cd <root_dir>
- _./build-ffmpeg.zsh_

- It shows the error:

  [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No


    =>   Checking for ccache...
    =>   No ccache found on the system
setup_target:9: project_root: parameter not set


- Try another command: _./build-ffmpeg.zsh windows-arm64_
It shows:

  [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No
  Dependencies: windows-arm64

    =>   Checking for ccache...
    =>   No ccache found on the system
setup_target:9: project_root: parameter not set


May I know the correct setup and the correct command to build? Thank you in advance!

PatTheMav:
> Hi, I'm try testing this pull request but failed to build. Please help to guide me. Here is the detailed info:
> 
> 1. Environment
> 
> * ARM64-based device
> * Windows 11 Pro 22H2
> * MSYS2, zsh, and ohmyzsh installed
> * MSVC toolchain (visual studio 2022, etc.)
> 
> 2. build
> 
> * clone the source code: _git clone https://github.com/PatTheMav/obs-deps.git -b ffmpeg-on-windows_
> * Open MSYS2 MSYS, which run zsh shell by default
> * cd <root_dir>
> * _./build-ffmpeg.zsh_
> * It shows the error:
>   [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No
>   =>   Checking for ccache...
>   =>   No ccache found on the system
>   setup_target:9: project_root: parameter not set
> * Try another command: _./build-ffmpeg.zsh windows-arm64_
>   It shows:
>   [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No
>   Dependencies: windows-arm64
>   =>   Checking for ccache...
>   =>   No ccache found on the system
>   setup_target:9: project_root: parameter not set
> 
> May I know the correct setup and the correct command to build? Thank you in advance!

The ZSH build scripts are for macOS only - for Windows the Powershell scripts need to be used.

qcvndemo01:
> > Hi, I'm try testing this pull request but failed to build. Please help to guide me. Here is the detailed info:
> > 
> > 1. Environment
> > 
> > 
> > * ARM64-based device
> > * Windows 11 Pro 22H2
> > * MSYS2, zsh, and ohmyzsh installed
> > * MSVC toolchain (visual studio 2022, etc.)
> > 
> > 
> > 2. build
> > 
> > 
> > * clone the source code: _git clone https://github.com/PatTheMav/obs-deps.git -b ffmpeg-on-windows_
> > * Open MSYS2 MSYS, which run zsh shell by default
> > * cd <root_dir>
> > * _./build-ffmpeg.zsh_
> > * It shows the error:
> >   [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No
> >   =>   Checking for ccache...
> >   =>   No ccache found on the system
> >   setup_target:9: project_root: parameter not set
> > * Try another command: _./build-ffmpeg.zsh windows-arm64_
> >   It shows:
> >   [OBS-DEPS] - configuration Release, target msys_nt-10.0-22621-x86_64, shared libraries: No
> >   Dependencies: windows-arm64
> >   =>   Checking for ccache...
> >   =>   No ccache found on the system
> >   setup_target:9: project_root: parameter not set
> > 
> > May I know the correct setup and the correct command to build? Thank you in advance!
> 
> The ZSH build scripts are for macOS only - for Windows the Powershell scripts need to be used.

Thanks for your comment.
I think this pull request can be built and run natively on Windows on Arm64 devices.
That's why I use MSYS2 on Windows to run .zsh script to have native ARM64 output binary.
Is my understanding correct?

PatTheMav:
MSYS2 is only needed and used for some dependencies that rely on a POSIX-like environment (`FFmpeg`, `x264`, and `vpx` are some examples), but the build system itself runs natively in PowerShell and invokes MSYS2 automatically when necessary.

The ZSH scripts are just for macOS.

qcvndemo01:
> MSYS2 is only needed and used for some dependencies that rely on a POSIX-like environment (`FFmpeg`, `x264`, and `vpx` are some examples), but the build system itself runs natively in PowerShell and invokes MSYS2 automatically when necessary.
> 
> The ZSH scripts are just for macOS.

Thank you, ZSH scripts are for MacOS, I got this point.

By using Build-Dependencies.ps1, the target will be x64 by default (".\Build-Dependencies.ps1 RelWithDebInfo ffmpeg" is the command I run).
May I know how to build ffmpeg and its dependencies natively on Windows on ARM64 device (so that I will be able to it natively on Windows on ARM64)?

Fenrirthviti:
Let's please move this conversation to either our Discord or Forums, as an open PR is not the correct place to receive support on how to build.

RytoEX (commented):
One thing I'm curious about is that the artifact names for FFmpeg Windows builds now seem to be `deps-ffmpeg-${arch}-${commitHash}`, whereas FFmpeg macOS builds are `ffmpeg-macos-${arch}-${commitHash}` (`-dSYMs` part optional). It would be best if we could choose one name pattern so that they are uniform and can be clearly identified.

I'll do a full review after other PRs land.

PatTheMav:
> One thing I'm curious about is that the artifact names for FFmpeg Windows builds now seem to be `deps-ffmpeg-${arch}-${commitHash}`, whereas FFmpeg macOS builds are `ffmpeg-macos-${arch}-${commitHash}` (`-dSYMs` part optional). It would be best if we could choose one name pattern so that they are uniform and can be clearly identified.
> 
> I'll do a full review after other PRs land.

The PR updates it to the correct pattern, the current pattern is "wrong" (as is the pattern for Qt) because of backwards compatibility with older Windows build scripts and CMake 2.0 build systems.

There are three separate packages:

* deps
* ffmpeg
* qt

So the correct naming is based on that - `deps-ffmpeg` and `deps-qt` is actually "wrong".

RytoEX:
> Hello, I am running into issues building this branch. Posting the discussion here: https://obsproject.com/forum/threads/obs-studio-on-windows-on-arm64-build-errors-on-ffmpeg-on-windows-branch.171207/
> > We are exploring building and using OBS Studio natively on Windows on ARM64 devices

We do not currently support Windows on Arm. That is a separate endeavor which will be addressed later. This PR is solely about moving our current FFmpeg builds from Linux to Windows.

tommyvct:
> Hello, I am running into issues building this branch. Posting the discussion here: https://obsproject.com/forum/threads/obs-studio-on-windows-on-arm64-build-errors-on-ffmpeg-on-windows-branch.171207/

That used to be possible, I have a PR draft and a fork for Windows ARM64. 

Unfortunately, that fork and PR has been in an abandoned state for quite some time now because I didn't have so much time in the past year. But now I may have some.

https://github.com/obsproject/obs-studio/wiki/Build-Instructions-for-Windows-ARM64
Here's my documentation on how to get the deps compiled, it may be outdated and some deps' linkage isn't the same as proposed here(e.g. `mbedtls`). Given that if the build system and ABI didn't changed dramatically in the last year, this doc may still be helpful.

Note that a lot of the deps here are cross-compiled from clang, which the toolchain's compatibility isn't as verified as mingw-gcc. Mixing toolchains together are usually not a recommended practice, and OBS Studio is using mixed toolchains since the very beginning. This PR is the chance to rectify this issue. It's also the prerequisite of getting Windows ARM64 officially supported.

deps.ffmpeg/30-svt-av1.ps1
```diff
@@ -0,0 +1,97 @@
...
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
+        '-DBUILD_APPS:BOOL=OFF'
+        '-DBUILD_DEC:BOOL=ON'
+        '-DBUILD_ENC=ON'
```
RytoEX, line 51:
```suggestion
        '-DBUILD_ENC:BOOL=ON'
```

deps.ffmpeg/40-aom.ps1
```diff
@@ -0,0 +1,102 @@
...
+
+    $Options = @(
+        $CmakeOptions
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
+        '-DENABLE_DOCS:BOOL=OFF'
+        '-DENABLE_EXAMPLES:BOOL=OFF'
+        '-DENABLE_TESTDATA:BOOL=OFF'
+        '-DENABLE_TESTS:BOOL=OFF'
+        '-DENABLE_TOOLS:BOOL=OFF'
+        "-DAOM_TARGET_CPU=$($TargetCPUs[$Target])"
```
RytoEX, line 48: Do we not also want:
```
-DENABLE_NASM:BOOL=ON
```

deps.ffmpeg/30-libvpx.ps1
```diff
@@ -0,0 +1,123 @@
...
+        BasePath = (Get-Location | Convert-Path)
+        BuildPath = "build_${Target}"
+        BuildCommand = $($ConfigureCommand -join ' ')
+        Target = $Target
+    }
+
+    $Backup = @{
+        MSYS2_PATH_TYPE = $env:MSYS2_PATH_TYPE
+    }
+    $env:MSYS2_PATH_TYPE='inherit'
```
RytoEX, line 70:
```suggestion
    $env:MSYS2_PATH_TYPE = 'inherit'
```

deps.ffmpeg/30-libvpx.ps1
```diff
@@ -0,0 +1,123 @@
...
+        BuildPath = "build_${Target}"
+        BuildCommand = "make -j${env:NUMBER_OF_PROCESSORS}"
+        Target = $Target
+    }
+
+    $Backup = @{
+        MSYS2_PATH_TYPE = $env:MSYS2_PATH_TYPE
+        VERBOSE = $env:VERBOSE
+    }
+    $env:MSYS2_PATH_TYPE='inherit'
```
RytoEX, line 90:
```suggestion
    $env:MSYS2_PATH_TYPE = 'inherit'
```

deps.ffmpeg/30-libvpx.ps1
```diff
@@ -0,0 +1,123 @@
...
+        BuildPath = "build_${Target}"
+        BuildCommand = "make install"
+        Target = $Target
+    }
+
+    $Backup = @{
+        MSYS2_PATH_TYPE = $env:MSYS2_PATH_TYPE
+        VERBOSE = $env:VERBOSE
+    }
+    $env:MSYS2_PATH_TYPE='inherit'
```
RytoEX, line 111:
```suggestion
    $env:MSYS2_PATH_TYPE = 'inherit'
```

deps.ffmpeg/10-zlib.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        '-DZ_HAVE_UNISTD_H:BOOL=OFF'
+    )
```
RytoEX, line 44: I suppose we don't need these if we aren't building this shared:
```
-DCMAKE_SHARED_LIBRARY_PREFIX=""
-DCMAKE_SHARED_LIBRARY_PREFIX_C=""
```

Though I don't recall why we build shared presently and why this PR no longer does. If we're not building this shared, we also don't need the `$OnOff` variable.

deps.ffmpeg/10-zlib.ps1
```diff
@@ -0,0 +1,87 @@
...
+        $Options += '--verbose'
+    }
+
+    $Options += @(
+        '--'
+        '/consoleLoggerParameters:Summary'
+        '/noLogo'
+        '/p:UseMultiToolTask=true'
+        '/p:EnforceProcessCountAcrossBuilds=true'
+    )
```
RytoEX, line 68: I kind of wish we had a "DefaultBuildOptions" util function that would remove the need for repeating this boilerplate everywhere, but I don't know if that would be problematic in other ways.

RytoEX (changes requested):
Submitting a preliminary review before I forget to do so. Initial testing seemed promising. Will investigate more soon.

utils.pwsh/Setup-Target.ps1
```diff
@@ -43,6 +43,44 @@ function Setup-BuildParameters {
...
+        }
+        MinSizeRel {
+            $script:CFlags += @(
+                '-O1 -Ob1 -DNDEBUG'
+            )
+            $script:CxxFlags += @(
+                '-O1 -Ob1 -DNDEBUG'
+            )
+        }
+    }
```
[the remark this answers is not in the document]
PatTheMav: Some project overwrite the CMake defaults with their own (bad) variations iirc, this enforces the defaults again. Annoying but necessary because some projects seem to have converted their autotools-based systems straight to CMake without regard for its defaults.

deps.ffmpeg/50-libtheora.ps1
```diff
@@ -0,0 +1,88 @@
...
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
+        "-DCMAKE_C_FLAGS=-wd4700"
```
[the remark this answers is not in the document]
PatTheMav: That was just a warning that was easy enough to disable (it's a warning about unused local variables).

deps.ffmpeg/40-aom.ps1
PatTheMav: It's omitted because it's supposed to use YASM.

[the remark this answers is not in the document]
PatTheMav: It's not because `CMAKE_SYSTEM_PROCESSOR` will not be aware of the target build architecture, it would always default to the host architecture.

Google also doesn't ship toolchain files for MSVC, so you need to enforce this manually.

See also: https://gitlab.kitware.com/cmake/cmake/-/issues/15170

utils.pwsh/Setup-Target.ps1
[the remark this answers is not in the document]
RytoEX: Unfortunate. I don't like having to compensate and maintain stuff like this in perpetuity, but I'm not sure it's easily avoidable here. We should attempt to see where we can alleviate this in our dependencies.

deps.ffmpeg/40-aom.ps1
RytoEX: It can use either.
https://aomedia.googlesource.com/aom/+/refs/tags/v3.8.0/build/cmake/aom_config_defaults.cmake#188
https://aomedia.googlesource.com/aom/+/refs/tags/v3.8.0/build/cmake/aom_configure.cmake#157

While both projects are still active on GitHub, NASM tags/releases more frequently. The current Yasm tag/release (1.3.0) is from 2014, while NASM's current release (2.16.01) is from 2022.

We also currently set this option:
https://github.com/obsproject/obs-deps/blob/e1d4ddd710dbcc08407ff4c5c6267372b6bb0c83/deps.ffmpeg/40-aom.zsh#L63

PatTheMav: I haven't retried this recently (the PR is fairly old by this point) but if I were to hazard a guess then I might have tried building with NASM and ran into issues with that precisely for the same reason as the change above (mainly that building with MSVC doesn't seem well supported).

PatTheMav:
```
    33 Warning(s)
    0 Error(s)
```
Seems like whatever might have been an issue isn't one anymore.

deps.ffmpeg/30-svt-av1.ps1
```diff
@@ -0,0 +1,97 @@
...
+        $Options += '--verbose'
+    }
+
+    $Options += @(
+        '--'
+        '/consoleLoggerParameters:Summary'
+        '/noLogo'
+        '/p:UseMultiToolTask=true'
+        '/p:EnforceProcessCountAcrossBuilds=true'
+    )
```
RytoEX, line 78:
```suggestion
    $Options += @($CmakePostfix)
```

deps.ffmpeg/10-zlib.ps1
PatTheMav: In general we don't seem to need it because when using the MSVC generator, their CMake detects that we're on windows and it omits the `lib` prefix by itself.

deps.ffmpeg/99-ffmpeg.ps1
```diff
@@ -0,0 +1,161 @@
+param(
+    [string] $Name = 'FFmpeg',
+    [string] $Version = '6.0',
+    [string] $Uri = 'https://github.com/FFmpeg/FFmpeg.git',
+    [string] $Hash = "a6dc92968a325d331bb6dcf9b3b2248026cd1d6c",
+    [array] $Patches = @(
+        @{
+            PatchFile = "${PSScriptRoot}/patches/FFmpeg/0001-flvdev-handle-unknown-Windows.patch"
```
RytoEX, line 8: flvdec*

RytoEX (changes requested):

deps.ffmpeg/40-aom.ps1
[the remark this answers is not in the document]
RytoEX: Unfortunate, but I understand now.

deps.ffmpeg/99-ffmpeg.ps1
```diff
@@ -0,0 +1,161 @@
...
+        '--enable-libsrt'
+        '--enable-shared'
+        '--enable-zlib'
+        '--disable-static'
+        '--disable-libjack'
+        '--disable-indev=jack'
+        '--disable-sdl2'
+        '--disable-doc'
+        '--disable-postproc'
+        '--disable-mediafoundation'
```
derrod, line 91: Is this still necessary? We might want to use it in the future for WoA and it's also usable for hardware decoding.

PatTheMav: We definitely need it for ARM64, but for now I preferred to keep the scope to "migrate existing builds to Windows" without bigger changes to the overall configuration (except where it was necessary to get it to compile at all).

tommyvct: We could add a if not Windows arm64 then add this config.

PatTheMav: That is out of scope of this PR - #190 exists for that.

RytoEX: As already mentioned by @PatTheMav , I'd prefer if the PR scope is kept minimal.

RytoEX:
#### Good news

OBS builds in Debug and RelWithDebInfo.

These worked:
* Encoders
  * AOM
  * SVT-AV1
  * AMF
  * NVENC
  * QSV
  * x264
* Recording
  * Standard
    * MKV
    * MP4
    * Fragmented MP4
  * Custom FFmpeg Output
    * MKV + libx264 + libvorbis
    * webm + libvpx-vp9 + libopus
* Streaming
  * RTMP
  * RTMPS
  * SRT (local network)
  * RIST (local network)
* Sources
  * Browser
    * Default
    * FishGL/Aquarium
  * Color
  * Display Capture
    * WGC
  * Game Capture
  * Image
    * Animated GIF
    * JPG
    * PNG
  * Image Slide Show
  * Media
    * FLV
    * MP4
    * SRT ingest on local network (OBS to OBS)
    * RIST ingest on local network (OBS to OBS)
    * webm with transparency
  * Text (FreeType 2)
  * Text (GDI+)
  * Video Capture Device
  * VLC
    * playlist of files
  * Window Capture
    * BitBlt
    * WGC
 * Filters
   * Chroma Key
   * Color Correction
 * webm stingers
 * Scripting
   * Python
   * Lua

Everything in my typical tests for those features worked.

#### Bad News

HLS streaming does not work. FFmpeg was compiled without HTTPS because the [schannel test](https://github.com/FFmpeg/FFmpeg/blob/a6dc92968a325d331bb6dcf9b3b2248026cd1d6c/configure#L6846-L6850) failed. Specifically, the `check_func_headers` test fails with:
```
error C2065: 'InitializeSecurityContext': undeclared identifier
```
I've also confirmed that this happens when trying to build FFmpeg locally. This seems to be caused by `-DWINAPI_FAMILY=WINAPI_FAMILY_APP` in the resulting defines for the test.

RytoEX:
After further investigation, `WINAPI_FAMILY_APP` is deprecated anyway. Setting `WINAPI_FAMILY` to `WINAPI_FAMILY_PC_APP` does not work. Setting it to `WINAPI_FAMILY_DESKTOP_APP` works fine. Also, not setting it causes it to default to `WINAPI_FAMILY_DESKTOP_APP`, which also works.

@PatTheMav 's recent change removes the define entirely, so it will default to `WINAPI_FAMILY_DESKTOP_APP` with current Win10/11 SDK headers. The CI log indicates that securechannel and HTTPS are now enabled. Will re-test HLS shortly.

RytoEX:
Retested HLS with the new changes and HLS now works, which means all of my tests passed.

Worth noting that some files may have different names (libsrt.dll -> srt.dll) which we may have to address via the updater, but that should be doable.

RytoEX (approved):

RytoEX:
For posterity, these files are renamed:
* bin/libsrt.dll -> bin/srt.dll
* lib/libsrt.lib -> lib/srt.lib
* lib/libmbedcrypto.lib -> lib/mbedcrypto.lib
* lib/libmbedtls.lib -> lib/mbedtls.lib
* lib/libmbedx509.lib -> lib/mbedx509.lib

These files have been removed:
* bin/libmbedcrypto.dll
* bin/libmbedtls.dll
* bin/libmbedx509.dll
* lib/zlibstatic.lib

I believe these changes are intentional.

PR #190, opened 2023-06-05
Add native arm64 build slice to all Windows dependencies

PatTheMav:
### Description
This is an exploratory PR for possible future native ARM64 builds on Windows.

Some dependencies cannot be built for ARM64 yet, others require specific workarounds, yet others can be built for ARM64 but themselves are too recent to be supported by OBS Studio.

This PR will stay in draft state until these and other issues are solved and are in a state that do not result in unnecessary support burden to current maintainers and supporters.

This PR is based on https://github.com/obsproject/obs-deps/pull/186 and the PRs it depends on.

### Motivation and Context
Due to Visual Studio 17 2022 having full support for ARM64 builds as well as `clang`- and `Ninja`-based builds, selecting the correct build architecture for MSVC builds does most of the work to achieve native builds, there are some caveats though:

* `libvpx` requires disabling more advanced NEON features not made available by Microsoft's ARM64 implementation.
* `x264` requires an update of the `gas-preprocessor` it ships with - luckily `FFmpeg` provides a very recent version of it (the version shipped with `x264` is 9 years old)
* `FFmpeg` requires the same `gas-processor` to be made available in the `PATH`
* `svt-av1` has not been made available as it's not well optimised for ARM64 - `dav1d` would be a better alternative, but `aom` is "good enough" for now
* `Qt6` builds without any additional hacks or issues
* `LuaJIT` has merged ARM64 and is available now
* `Python` supports Windows ARM64 since version 3.11, alas OBS Studio does not support Python 3.11 at runtime yet (you could build OBS Studio with it, but people couldn't add its libraries at runtime)
* `mbedtls` is still **broken** on ARM64 with version 3.5.1
* `libdatachannel` requires at least `mbedtls` 3.4.0, so both are entirely broken on ARM64 right now
* `vpl` is disabled for ARM64
* `Aja NTV2` is disabled for ARM64

### How Has This Been Tested?
Tested on Windows 11 ARM64.

### Types of changes

### Checklist:


driver1998:
What is the error exactly in building mbedtls? I saw upstream fixed similar issues multiple times.

PatTheMav:
> What is the error exactly in building mbedtls? I saw upstream fixed similar issues multiple times.

Compiling `aes.c` breaks with version 3.5.1:

```
error C1189: #error:  This header is specific to X86, X64, ARM64, and ARM64EC targets
```

This is possibly fixed by https://github.com/Mbed-TLS/mbedtls/issues/8168, so we're waiting for MbedTLS to cut a new release with those fixes included.

tommyvct:
This needs to be rebased and merge #227 in.

utils.pwsh/Install-BuildDependencies.ps1
tommyvct, line 1: Small nitpicks here. If you look a few lines above:

```pwsh
    $Prefixes = @{
        'arm64' = ${env:ProgramFiles(arm)}
        'x64' = ${env:ProgramFiles}
        'x86' = ${env:ProgramFiles(x86)}
    }
```

This is actually incorrect. `C:\Program Files` is reserved for the programs that is native to this machine, or at least the same bitness. For example, an x64 version of vscode machine-wide installer will install vscode in `C:\Program Files`. In another word, both x64 and arm64 programs sits in `C:\Program Files`.

`${env:ProgramFiles(arm)}` is a different story, usually it maps to `C:\Program Files (Arm)` which is usually a place for arm32 programs, which is extremely rare to see these days.

RytoEX:
> > together off-thread. Rest assured, that we discuss this regularly.
> 
> May I join the conversation?

I appreciate the enthusiasm, but these are conversations that we have either on voice calls or in ongoing DMs about getting stuff to work. It is mostly stuff like "so about that PR" and "here's an idea" and "here's how we might get that to work when we have time". The point is that he doesn't need a reminder to rebase this. He's aware. This is on our roadmap to address.

tommyvct:
How about we start a separate thread in the development channel? Or I could just ask and @ both of you in the channel?

deps.qt/qt6.ps1
```diff
@@ -143,6 +143,7 @@ function Configure {
     $CMakeTarget = @{
         x64 = 'x64'
         x86 = 'Win32'
+        arm64 = 'arm64'
     }
 
     $Options = ($Options -join ' ') -replace '-G Visual Studio \d+ \d+','-G Ninja' -replace "-A $($CMakeTarget[$Target])",''
```
RytoEX, line 149: Qt requires a working Qt host with precompiled tools (moc, uic, etc.). Here's a quick and dirty hack that got me working locally. We would probably have to have the arm64 job wait for a successful x64 run, download the build, extract it somewhere, and point the arm64 job at that Qt directory.
```diff
diff --git a/deps.qt/qt6.ps1 b/deps.qt/qt6.ps1
index a7aea57..f9ee2a9 100644
--- a/deps.qt/qt6.ps1
+++ b/deps.qt/qt6.ps1
@@ -146,6 +146,13 @@ function Configure {
         arm64 = 'arm64'
     }

+    if ( $Target -eq "arm64" ) {
+        $Options += @(
+            '-DQT_QMAKE_TARGET_MKSPEC:STRING=win32-arm64-msvc'
+            '-DQT_HOST_PATH:PATH=C:\Qt\6.6.1\msvc2019_64'
+        )
+    }
+
```

deps.windows/60-python.ps1
```diff
@@ -2,10 +2,11 @@ param(
     [string] $Name = 'python',
     [string] $Version = '3.1.1',
     [string] $Uri = 'https://github.com/pyenv-win/pyenv-win.git',
-    [string] $Hash = '754a6ca877f66aaa2bd4485a29411267d0705273',
+    [string] $Hash = '3201b6c2bf1e4d4791f66fc8634e80f3ff2731c6',
```
RytoEX, line 5: Is this just to get a newer pyenv-win cache?

deps.windows/40-detours.ps1
```diff
@@ -2,26 +2,26 @@ param(
     [string] $Name = 'detours',
     [string] $Version = '4.0.1',
     [string] $Uri = 'https://github.com/microsoft/detours.git',
-    [string] $Hash = 'e4bfd6b03e50de46b47abfbd1e46b384f0c5f833'
+    [string] $Hash = '734ac64899c44933151c1335f6ef54a590219221'
```
RytoEX, line 5: Is there a specific change needed between these hashes? Last time I asked about updating this, I was asked/told not to update from the stable tag, old as it may be.

deps.windows/40-luajit.ps1
```diff
@@ -2,11 +2,15 @@ param(
     [string] $Name = 'luajit',
     [string] $Version = '2.1',
     [string] $Uri = 'https://github.com/luajit/luajit.git',
-    [string] $Hash = '8635cbabf3094c4d8bd00578c7d812bea87bb2d3'
+    [hashtable] $Hashes = @{
+        x64 = '8635cbabf3094c4d8bd00578c7d812bea87bb2d3'
+        x86 = '8635cbabf3094c4d8bd00578c7d812bea87bb2d3'
+        arm64 = '43d0a19158ceabaa51b0462c1ebc97612b420a2e'
+    }
```
RytoEX, line 9: Will be obsoleted once we update this for all systems.

utils.pwsh/Setup-Target.ps1
```diff
@@ -29,13 +46,6 @@ function Setup-BuildParameters {
         . $PSScriptRoot/Logger.ps1
     }
 
-    $NumProcessors = (Get-CimInstance Win32_ComputerSystem).NumberOfLogicalProcessors
-
-    if ( $NumProcessors -gt 1 ) {
-        $env:UseMultiToolTask = $true
-        $env:EnforceProcessCountAcrossBuilds = $true
-    }
-
```
RytoEX, line 38: Was this moved somewhere, or is it no longer needed?

deps.ffmpeg/99-ffmpeg.ps1
```diff
@@ -130,11 +134,13 @@ function Build {
     }
 
     $Backup = @{
-        MSYS2_PATH_TYPE = $env:MSYS2_PATH_TYPE
+        MSYS2_PATH_TYPE = $Env:MSYS2_PATH_TYPE
```
RytoEX, line 137: Stray `Env`.
```suggestion
        MSYS2_PATH_TYPE = $env:MSYS2_PATH_TYPE
```

deps.ffmpeg/30-libvpx.ps1
```diff
@@ -1,12 +1,12 @@
 param(
     [string] $Name = 'libvpx',
     [string] $Version = '1.13.1',
-    [string] $Uri = 'https://github.com/webmproject/libvpx/archive/refs/tags/v1.13.1.zip',
-    [string] $Hash = "${PSScriptRoot}/checksums/v1.13.1.zip.sha256"
+    [string] $Uri = 'https://github.com/webmproject/libvpx.git',
+    [string] $Hash = '1672f4db71370e07e43e1e121cf2e92bf97ea4ff'
```
RytoEX, line 5: v1.14.0-rc1 is out, so hopefully there's a stable tag that can be used soon.

RytoEX (changes requested):
Just some quick observations. I did get a local Qt build working, but it requires setting both `QT_QMAKE_TARGET_MKSPEC` and `QT_HOST_PATH`. I've commented how I might approach this for CI. We can attempt to optimize it later.

deps.windows/40-detours.ps1
PatTheMav: Only the most recent tags support ARM64 at all in my tests.

deps.windows/60-python.ps1
PatTheMav: That's required to get a Pyenv version that supports Python 3.11 and Windows on ARM at all.

RytoEX: You can also (or should be able to) refresh the cache if it's just about updating the cache of available Python versions. If there's something specific to the pyenv-win code itself though, that's worth noting.

PatTheMav: Afaik the Pyenv version itself didn't have support for it.

deps.ffmpeg/30-libvpx.ps1
shijunz: I am trying the PR, but i get many errors for example:
1 git clone :  https://github.com/PatTheMav/obs-deps/tree/windows-arm64-native, and checkout windows-arm64-native  branch,
2 try to run it locally:  .\Build-Dependencies.ps1 -Target 'arm64' -Configuration 'Debug' -PackageName 'ffmpeg' 

3 many libraries are built successfully, but some libraries fail, for example, mbedtls,

   1>Checking Build System
  Building Custom Rule C:/tools/alpaca/obs/PatTheMav-obs-deps/windows_build_temp/mbedtls/library/CMakeLists.txt
  asn1parse.c
  aesce.c
  aesni.c
  aria.c
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.36.32532\include\immintrin.h(15,1): fatal error C1189: #error:  This header is spe
cific to X86, X64, ARM64, and ARM64EC targets [C:\tools\alpaca\obs\PatTheMav-obs-deps\windows_build_temp\mbedtls\build_arm64\library\mbedcrypto.vcxproj]
  base64.c


another questions, I have no ideay how to build these modules

•	libvpx: Uses shell support and make from MSYS2, compiled with MSBuild
•	x264: Uses shell support and make from MSYS2, compiled with cl
•	lame: Built natively using old (but still working) nmake script


@PatTheMav , could you guide me ?

thanks

PatTheMav: https://github.com/obsproject/obs-deps/pull/190#issuecomment-1886167857

deps.windows/40-detours.ps1
RytoEX: FWIW, we _are_ using the most recent tag, but there have been new commits since then.

deps.windows/40-luajit.ps1
RytoEX: Should be obsoleted by https://github.com/obsproject/obs-deps/commit/96549337dff09177ff75ed480a629b5cb5e7c97f and https://github.com/obsproject/obs-deps/commit/ec04a292eb77c141ac5ae603b5ec92121366b17c (#228), which updated to a newer LuaJIT commit.

deps.windows/60-python.ps1
RytoEX: This is probably fine. While I'd prefer that pyenv-win had a new tag, not much we can do there.

RytoEX:
Was just skimming back over this.

> * `x264` requires an update of the `gas-preprocessor` it ships with - luckily `FFmpeg` provides a very recent version of it (the version shipped with `x264` is 9 years old)
> * `FFmpeg` requires the same `gas-processor` to be made available in the `PATH`

This seems like it should be a trivial Merge Request on the x264 repo. Ideally, I'd like the change to happen there at some point.

> * `svt-av1` has not been made available as it's not well optimised for ARM64 - `dav1d` would be a better alternative, but `aom` is "good enough" for now

VideoLAN's dav1d is a decoder only. It is not an encoder. Xiph's rav1e is an encoder.

deps.windows/40-detours.ps1
RytoEX: Until a new stable tag emerges, we're going to continue using the stable tag for x64/x86, but a recent commit for arm64.

PatTheMav:
With the last commit, all dependencies also build for Windows ARM64 now. Remaining tasks:

* Update the scheduled workflow
* Test Windows artefacts on x64 and ARM64
* Update OBS-Studio to accommodate MbedTLS 3.6.0
* Update OBS-Studio to build for ARM64

megabytesme:
Hey! I'd be down to test this (noticed the "Seeking Testers" tag). Is there a release I could download and test? Not used to GitHub haha.

PatTheMav:
> As an aside, I noticed that this PR adds `deps.ffmpeg/patches/libpng/0001-enable-ARM-NEON-optimisations-windows.patch`, but this patch already exists at `deps.ffmpeg/patches/libpng/0001-enable-ARM-NEON-optimisations.patch`, though it is not currently used in the libpng script.

That's because of Windows line endings iirc.

RytoEX:
> Was just skimming back over this.
> 
> > * `x264` requires an update of the `gas-preprocessor` it ships with - luckily `FFmpeg` provides a very recent version of it (the version shipped with `x264` is 9 years old)
> > * `FFmpeg` requires the same `gas-processor` to be made available in the `PATH`
> 
> This seems like it should be a trivial Merge Request on the x264 repo. Ideally, I'd like the change to happen there at some point.

Just figured I'd check. The version of `gas-preprocessor` in the x264 repo is identical to this:
https://github.com/FFmpeg/gas-preprocessor/blob/ee12830747ff0b97ec6b41f4263fec63d1711365/gas-preprocessor.pl

It might be worth us at least opening an Issue on the x264 repo pointing out that it's outdated and does not cross-compile on Windows as-is.

That said, it would be nice to be able to submit a Merge Request to x264 that simply updates `gas-preprocessor`, but I'm not clear on their CLA requirements.

RytoEX:
> > As an aside, I noticed that this PR adds `deps.ffmpeg/patches/libpng/0001-enable-ARM-NEON-optimisations-windows.patch`, but this patch already exists at `deps.ffmpeg/patches/libpng/0001-enable-ARM-NEON-optimisations.patch`, though it is not currently used in the libpng script.
> 
> That's because of Windows line endings iirc.

Hmm. I see. A little unfortunate. We should look into submitting an Issue or PR so we can stop maintaining this patch in three places.

anthony-linaro:
Hi all, just poking my head in here - what more needs to be done for this? What is the user testing that needs to be done?

I maintain the Blender Windows ARM64 builds, so have a little experience in this area, as we build ffmpeg as part of that too (although for Blender, given encoding speed isn't majorly critical compared to other bits like rendering, I just switched ASM off for x264 on these platforms...)

PatTheMav:
> Hi all, just poking my head in here - what more needs to be done for this? What is the user testing that needs to be done?
> 
> I maintain the Blender Windows ARM64 builds, so have a little experience in this area, as we build ffmpeg as part of that too (although for Blender, given encoding speed isn't majorly critical compared to other bits like rendering, I just switched ASM off for x264 on these platforms...)

As far as only `obs-deps` is concerned, they are done. 

The next major part is updating `obs-studio`'s build system to support ARM64 on Windows, which requires some extensive rewrites as the build system implicitly only knows Windows as an AMD64 platform, which also requires support for 32-bit x86 applications (e.g. our capture hook needs to be built and shipped as x86 as well to support capturing older games).

As Windows ARM64 supports running x64 as well as x86 binaries (ARM32 as well, but we're not concerned about those), it effectively means that users might be running apps and games with those architectures as well. Thus an ARM64 build needs x64 and x86 variants of the capture hooks and virtual camera, which adds complexity.

The main issue is that CMake does not support generating multi-architecture Visual Studio projects (i.e., a project that would have variants of the capture hook solution targeted for all 3 architectures) - each project can target only a specific architecture.

So if we want to build OBS Studio for ARM64, we need to generate a "full" ARM64 project and two cut-down projects for x64 and x86 respectively and trigger them from the ARM64 project and also collect all built artefacts to generate the final shippable product.

Effectively we have to write _a lot_ of CMake code around CMake's own limitations in relation to Visual Studio projects as there hasn't been much momentum or desire to support this scenario (because our expectation is that one only has to generate an ARM64 project of `obs-studio` and the x86 and x64 parts will automatically be generated and built as well).

---

Once OBS Studio has been updated to support build ARM64 to begin with, we can actually _test_ the builds generated by this PR. Only after identifying and fixing any issues that might arise from that can we consider this PR to be ready for review.

tommyvct:
> The next major part is updating `obs-studio`'s build system to support ARM64 on Windows, which requires some extensive rewrites as the build system implicitly only knows Windows as an AMD64 platform, which also requires support for 32-bit x86 applications (e.g. our capture hook needs to be built and shipped as x86 as well to support capturing older games).

This is specific to Game Capture, right? I guess for now we could just let this slide for now, since the last known good ARM64 build works perfectly fine on Desktop capture, and that would cover at least 90% of the use cases. Snapdragon X Elite/Plus isn't intended to be a video game platform, so this could be put at the bottom of the list.

RytoEX:
> > The next major part is updating `obs-studio`'s build system to support ARM64 on Windows, which requires some extensive rewrites as the build system implicitly only knows Windows as an AMD64 platform, which also requires support for 32-bit x86 applications (e.g. our capture hook needs to be built and shipped as x86 as well to support capturing older games).
> 
> This is specific to Game Capture, right? I guess for now we could just let this slide for now, since the last known good ARM64 build works perfectly fine on Desktop capture, and that would cover at least 90% of the use cases. Snapdragon X Elite/Plus isn't intended to be a video game platform, so this could be put at the bottom of the list.

Game Capture and Virtual Camera both require both 64-bit and 32-bit components on x64 Windows. The minute we ship a build without features we'll be asked why it doesn't include X, Y, and Z repeatedly. Given that historically, we have only had to contend with x86 (32-bit) and x64 (64-bit) on Windows, there are components that include 32 or 64 or the like in their name, so the build specifications for these (in CMake) and code that interfaces with those filenames must be made aware that a non-32-bit, non-64-bit type/name is possible (arm).

Rest assured, we have considered these points when thinking about our path forward.

tommyvct:
Well, we could release a alpha version that lacks that feature, put a huge dialog with huge font, that is non-dismissible for 30 seconds, saying that by continue using you agree to acknowledge that feature X, Y, and Z are missing as they are being worked on, and please don't ask about these features anywhere.

The problem I see was Qualcomm used OBS Studio to promote their Snapdragon X Elite platform, while we have no support for Windows on ARM at all, and absolutely no contribution from Qualcomm themselves.

RytoEX:
> Well, we could release a alpha version that lacks that feature, put a huge dialog with huge font, that is non-dismissible for 30 seconds, saying that by continue using you agree to acknowledge that feature X, Y, and Z are missing as they are being worked on, and please don't ask about these features anywhere.

Thanks for the suggestion, but we already have our own plans for the path forward.

> The problem I see was Qualcomm used OBS Studio to promote their Snapdragon X Elite platform, while we have no support for Windows on ARM at all, and absolutely no contribution from Qualcomm themselves.

We're aware.

deps.qt/qt6.ps1
```diff
@@ -122,9 +122,19 @@ function Configure {
         '-DCMAKE_IGNORE_PREFIX_PATH:PATH=C:/Strawberry/c'
     )
 
+    if ( $env:QtHostPath -ne $null ) {
+        $Options += @(
+            "-DQT_HOST_PATH=${env:QtHostPath}"
```
RytoEX, line 127:
```suggestion
            "-DQT_HOST_PATH:PATH=${env:QtHostPath}"
```
Should this have the type `PATH` specified?

deps.qt/qt6.ps1
```diff
@@ -206,9 +216,19 @@ function Qt-Add-Submodules {
         )
     }
 
+    if ( $env:QtHostPath -ne $null ) {
+        $Options += @(
+            "-DQT_HOST_PATH=${env:QtHostPath}"
```
RytoEX, line 221:
```suggestion
            "-DQT_HOST_PATH:PATH=${env:QtHostPath}"
```
Should this have the type `PATH` specified?

deps.qt/qt6.ps1
```diff
@@ -122,9 +122,19 @@ function Configure {
         '-DCMAKE_IGNORE_PREFIX_PATH:PATH=C:/Strawberry/c'
     )
 
+    if ( $env:QtHostPath -ne $null ) {
+        $Options += @(
+            "-DQT_HOST_PATH=${env:QtHostPath}"
+            '-DCMAKE_CROSSCOMPILING:BOOL=ON'
+            '-DQT_FORCE_BUILD_TOOLS:BOOL=ON'
+            '-DCMAKE_SYSTEM_NAME:STRING=Windows'
```
RytoEX, line 130: Sort these into `CMAKE_` and `QT_` prefixes.

deps.qt/qt6.ps1
```diff
@@ -206,9 +216,19 @@ function Qt-Add-Submodules {
         )
     }
 
+    if ( $env:QtHostPath -ne $null ) {
+        $Options += @(
+            "-DQT_HOST_PATH=${env:QtHostPath}"
+            '-DCMAKE_CROSSCOMPILING:BOOL=ON'
+            '-DQT_FORCE_BUILD_TOOLS:BOOL=ON'
+            '-DCMAKE_SYSTEM_NAME:STRING=Windows'
```
RytoEX, line 224: Sort these into `CMAKE_` and `QT_` prefixes.

RytoEX (changes requested):
Minor nits.

Drop the Sparkle commit. Sparkle is already updated in this repo.

Would still like to see these additional patches submitted to the dependency repos to see if they'll adopt them.

jagp-qcom:
Qualcomm is working on ARM native porting. Very soon we shall be more active here.

Fenrirthviti:
> Qualcomm is working on ARM native porting. Very soon we shall be more active here.

Qualcomm has not reached out to us in any capacity at this time. We hope that they would contact us if they are starting work on the project to port it to a new platform to ensure that efforts are aligned.

DoM1niC:
I got the new Surface X, if somebody is needed for Tests so feel free to link me a Build 🙂👍🏻

sameeraman:
I'm happy to test also if help needed on that. Keen to get a ARM64 windows version of the OBS.

RytoEX:
Any test builds would likely be provided on PRs on the [obs-studio repo](https://github.com/obsproject/obs-studio). In the meantime, I would like to keep comments focused on review or feedback from people who have actually tried to build/test these changes. Thank you for understanding.

talynone:
> As Windows ARM64 supports running x64 as well as x86 binaries (ARM32 as well, but we're not concerned about those), it effectively means that users might be running apps and games with those architectures as well. Thus an ARM64 build needs x64 and x86 variants of the capture hooks and virtual camera, which adds complexity.

Just a minor correction, Windows 11 does not support ARM32 on the new Snapdragon Elite processors, and there's no emulation layer for it either, they simply won't work. It's the reason Android subsystem for Windows broke on the Elite processors, it relied on ARM32 binaries.

utils.pwsh/Setup-Target.ps1
RytoEX: This feedback thread is still open.

PatTheMav: It's not needed anymore because the same is achieved via the automatic CMake postfix and for non-CMake based projects it's disabled because `make` triggers a single invocation of `cl.exe` for every source file anyway.

RytoEX: Thanks for explaining.

RytoEX:
As discussed off-thread with @PatTheMav , updating libsrt to 1.5.3 has known issues in OBS Studio on Windows. These are supposedly resolved in libsrt 1.5.4, which is not yet out/stable. It's not clear to me that libsrt 1.5.3+ is required for WoA/arm64. Can we confirm?

Separately, MbedTLS 3.6.0+ _is_ required for WoA/arm64, as far as I know. However, we cannot update that until https://github.com/Haivision/srt/issues/2957 is fixed. This seems to have been addressed by `deps.ffmpeg/patches/srt/0002-update-mbedtls-discovery-windows.patch` included in this PR. We should submit that as a PR to the srt repo.

If we _need_ libsrt 1.5.3+, either we wait for all dependencies to be stable, or we teach the Windows dependency build scripts to build different versions per target architecture.

Additionally, I still cannot build obs-studio on Windows against libsrt 1.5.3+ and mbedTLS 3.6.0+ due to the bcrypt linking issue, even with the patch here. I haven't tried https://github.com/obsproject/obs-studio/pull/11056 because it fails on CI, and needs to be rebased (I'm guessing gersemi changes).

sudopluto:
i just got my new arm thinkpad, will try to build & please ping me if any future testing is needed, will be happy to try : )

PatTheMav:
Updated srt to 1.5.3 and MbedTLS to 3.5.2.

RytoEX:
Can you fix the merge conflict here so that CI creates new artifacts?

PatTheMav:
> Can you fix the merge conflict here so that CI creates new artifacts?

Done.

deps.ffmpeg/30-libvpx.ps1
```diff
@@ -40,7 +41,7 @@ function Configure {
         '../configure'
         ('--prefix="' + $($script:ConfigData.OutputPath -replace '([A-Fa-f]):','/$1' -replace '\\','/') + '"')
         ('--target=' + $($BuildTargets[$Target]))
-        '--enable-runtime-cpu-detect'
```
RytoEX, line 43: Why is this removed?

deps.ffmpeg/40-aom.ps1
```diff
@@ -47,6 +48,7 @@ function Configure {
         '-DENABLE_TOOLS:BOOL=OFF'
         '-DENABLE_NASM:BOOL=ON'
         "-DAOM_TARGET_CPU=$($TargetCPUs[$Target])"
+        $(if ( $Target -eq 'arm64' ) { '-DCONFIG_RUNTIME_CPU_DETECT=0' })
```
RytoEX, line 51: Why is this added?

deps.ffmpeg/40-x264.ps1
```diff
@@ -3,17 +3,24 @@ param(
...
+        if ( $Target -ne 'arm64' ) {
+            Invoke-External pacman.exe -S --noconfirm --needed --noprogressbar nasm
+        }
         Invoke-External pacman.exe -S --noconfirm --needed --noprogressbar make
     }
+
+    if ( $Target -eq 'arm64' ) {
+        Remove-Item -Path "${Path}/tools/gas-preprocessor.pl" -ErrorAction SilentlyContinue
+        Copy-Item -Path "$($script:WorkRoot)/gas-preprocessor/gas-preprocessor.pl" -Destination "${Path}/tools/"
+    }
```
RytoEX, line 23: Would this be more appropriate in the `Patch()` function?

deps.ffmpeg/60-lame.ps1
```diff
@@ -19,10 +19,11 @@ function Setup {
...
     Set-Location "${Name}-${Version}"
-    if ( Test-Path "build_${Target}" ) {
-        Log-Information "Clean build directory (${Target})"
-        Remove-Item -Path "build_${Target}" -Recurse -Force
-    }
+
+    Get-ChildItem -Recurse -Include 'lame.exe','mp3x.exe','mp3rtp.exe' | Remove-Item
+    Get-ChildItem -Recurse -Include 'libmp3lame.*','libmp3lame-static.lib','lame_enc.dll' | Remove-Item
+    Get-ChildItem -Recurse -Include 'lame.pdb','icl.pch' | Remove-Item
+    Get-ChildItem -Recurse -Include '*.obj','*.res' | Remove-Item
```
RytoEX, line 26: Noting for posterity that this is the correct syntax due to a bug in `Remove-Item`:
* https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/remove-item?view=powershell-7.4#example-4-delete-files-in-subfolders-recursively

deps.ffmpeg/60-lame.ps1
```diff
@@ -42,14 +43,20 @@ function Build {
...
         x64 = 'x64'
         x86 = 'I686'
+        arm64 = 'arm64'
     }
 
     $Params = @{
         BasePath = (Get-Location | Convert-Path)
         BuildPath = "."
-        BuildCommand = "nmake -f Makefile.MSVC MACHINE=/machine:$($BuildMachines[$Target]) COMP=MS ASM=NO MSVCVER=Win64"
+        BuildCommand = "nmake -f Makefile.MSVC MACHINE=/machine:$($BuildMachines[$Target]) MMX=NO COMP=MS ASM=NO MSVCVER=Win64"
```
RytoEX, line 52: Why the additional `MMX=NO` here?

RytoEX (changes requested):
Some questions and some paper trail keeping.

deps.ffmpeg/30-libvpx.ps1
PatTheMav: There had been issues with it when I started porting to ARM64 on Windows, but I cannot reproduce any compile issues with it anymore. Maybe issues will exhibit themselves at runtime (as e.g. Google themselves always disable runtime detection for their own ARM64 builds on Android), but I haven't been able to run into any yet.

deps.ffmpeg/40-aom.ps1
PatTheMav: Like with vpx I had issues with this when porting, but it compiles fine with current Visual Studio and version 3.9.1. Issues might occur only during runtime as well.

deps.ffmpeg/40-x264.ps1
PatTheMav: Using a patch for this would remove the ability to automatically update the preprocessor with an updated version from its own git repository. This way we'd just have to update the gas-preprocessor commit hash.

deps.ffmpeg/60-lame.ps1
PatTheMav: It doesn't compile otherwise (as ARM64 has no MMX obviously) and the performance impact should be negligible on modern CPUs anyway.

deps.ffmpeg/40-x264.ps1
RytoEX: Fair.

deps.ffmpeg/60-lame.ps1
RytoEX: Fair.

RytoEX:
SRT streaming works with these deps and I cannot seem to reproduce any freezes or crashes, though I am admittedly not clear on the exact steps to reproduce the previously reported freeze/crash.

RTMPS streaming to YouTube works fine.

HLS streaming to YouTube _sometimes_ seems to fail with these deps. The failure state will result in the stream output dropping all frames. Clicking "Stop Streaming" will result in a stuck "Stopping stream..." button, and trying to exit will hang the application. I suspect the "Stopping" and hang issues are not related to these deps specifically, and I'm not sure that the initial failure is either (it could just be timing issues with YouTube's disconnect detection with HLS weirdness).

MbedTLS and SRT are the only deps updated here, so the tests above seem sufficient for now. If we find issues with either update, we'll need to aggressively debug them, or figure out a way to use the newer deps only for WoA.

RytoEX (approved):

Fritomaster:
Appreciate you guys working on this. As an artist who streams my work and uses mostly arm now this would be great!

i-lina:
Thank you for this feature. Which release is this feature targeted at?

PR #192, opened 2023-06-05
deps.qt: Add qtcharts

pkviet:
### Description
This enables building of qtcharts (as well as qtopengl, qtopenglwidgets) on which it depends.

### Motivation and Context
This is a companion to other PRs in obs-studio enabling stats charts for SRT [1] & maybe regular stats.

[1] https://github.com/obsproject/obs-studio/pull/9027

### How Has This Been Tested?
Builds locally on windows and passes CI on my fork.

### Types of changes

### Checklist:

### DIFF
--- a/deps.qt/qt6.ps1
+++ b/deps.qt/qt6.ps1
@@ -8,6 +8,7 @@ param(
 
 $QtComponents = @(
     'qtbase'
+    'qtcharts'
     'qtimageformats'
     'qtshadertools'
     'qtmultimedia'
@@ -110,6 +111,7 @@ function Configure {
         '-DFEATURE_system_zlib:BOOL=OFF'
         '-DFEATURE_testlib:BOOL=OFF'
         '-DFEATURE_windeployqt:BOOL=OFF'
+        '-DINPUT_opengl:STRING=no'
         '-DINPUT_openssl:STRING=no'
         '-DQT_BUILD_BENCHMARKS:BOOL=OFF'
         '-DQT_BUILD_EXAMPLES:BOOL=OFF'
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -14,6 +14,7 @@ local -a patches=(
 
 local -a qt_components=(
   'qtbase'
+  'qtcharts'
   'qtimageformats'
   'qtshadertools'
   'qtmultimedia'
@@ -131,6 +132,7 @@ config() {
     -DFEATURE_system_zlib:BOOL=ON
     -DFEATURE_testlib:BOOL=OFF
     -DFEATURE_windeployqt:BOOL=OFF
+    -DINPUT_opengl:STRING=no
     -DINPUT_openssl:STRING=no
     -DQT_BUILD_BENCHMARKS:BOOL=OFF
     -DQT_BUILD_EXAMPLES:BOOL=OFF
--- a/deps.qt/checksums/qtcharts-everywhere-src-6.8.3.tar.xz.sha256
+++ b/deps.qt/checksums/qtcharts-everywhere-src-6.8.3.tar.xz.sha256
@@ -1,0 +1,1 @@
+34aa5f5a48bb7481c9bd62172696831c73c2cef327aa6b871b1861693b3e0914  qtcharts-everywhere-src-6.8.3.tar.xz
--- a/deps.qt/checksums/qtcharts-everywhere-src-6.8.3.zip.sha256
+++ b/deps.qt/checksums/qtcharts-everywhere-src-6.8.3.zip.sha256
@@ -1,0 +1,14 @@
+<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04">
+  <Obj RefId="0">
+    <TN RefId="0">
+      <T>Microsoft.PowerShell.Commands.FileHashInfo</T>
+      <T>System.Object</T>
+    </TN>
+    <ToString>Microsoft.PowerShell.Commands.FileHashInfo</ToString>
+    <Props>
+      <S N="Algorithm">SHA256</S>
+      <S N="Hash">aaf40d44eb54ea8b349f6e409f9c4ebabdcabcacce1f4045a01133b17f92117e</S>
+      <S N="Path">qtcharts-everywhere-src-6.8.3.zip</S>
+    </Props>
+  </Obj>
+</Objs>


deps.qt/qt6.ps1
```diff
@@ -12,6 +12,7 @@ $QtComponents = @(
     'qtshadertools'
     'qtmultimedia'
     'qtsvg'
+    'qtcharts'
```
RytoEX, line 15: Does qtcharts require anything other than qtbase? If not, then it should be placed just below qtbase.

deps.qt/qt6.zsh
```diff
@@ -18,6 +18,7 @@ local -a qt_components=(
   'qtshadertools'
   'qtmultimedia'
   'qtsvg'
+  'qtcharts'
```
RytoEX, line 21: Does qtcharts require anything other than qtbase? If not, then it should be placed just below qtbase.

RytoEX (changes requested):
As far as I can tell, qtopengl and qtopenglwidgets are optional dependencies of qtcharts, not required dependencies. Could you verify?

deps.qt/qt6.ps1
pkviet: you're right : https://github.com/qt/qtcharts/blob/dev/dependencies.yaml
But for CI on linux, I noticed that CI failed if I didn't add qtopenGL and qtopenGLWidgets in CI/linux/01_install_dependencies.sh

RytoEX: That's strange. Though, that seems like a Linux-only concern, unless we can confirm that it actually builds these components on Windows or macOS.

pkviet: positioned qtcharts just below qtbase

deps.qt/qt6.zsh
pkviet: fixed

pkviet:
> As far as I can tell, qtopengl and qtopenglwidgets are optional dependencies of qtcharts, not required dependencies. Could you verify?

correct:
https://github.com/qt/qtcharts/blob/dev/src/charts/CMakeLists.txt#L123-L128

pkviet:
also:
https://github.com/qt/qtcharts/blob/dev/CMakeLists.txt#L19-L20

deps.qt/qt6.ps1
pkviet: > That's strange. Though, that seems like a Linux-only concern, unless we can confirm that it actually builds these components on Windows or macOS, so I'm not sure "and its dependencies" needs mentioned.

good point. I removed qtopengl and qtopenglwidgets but had forgotten to update the commit message to my utter shame 🤦

pkviet:
@RytoEX mmh even though qt6OpenGL is not mandatory it seems it builds with Qt6OpenGL and Qt6OpenGLWidgets on default even if the latter are not specified.
See the screenshot:
DependenciesGui_2023-07-08_06-49-18
I'll investigate if I do need OpenGL/OpenGLWidgets or not in my current code and how to force building without these deps.

pkviet:
It seems Qt6OpenGL and Qt6OpenGlWidgets are already shipping with the deps.
This is a screenshot from the last obs-deps release (06-22):
explorer_2023-07-08_07-01-35

I checked the various qt dlls in the bin folder but none had QtOpenGL as a dependency ...

RytoEX:
I do see that. We don't currently ship them with OBS Studio itself because they aren't needed. Qt's default configuration is "dynamic OpenGL". The old [Qt 5.15 build instructions](https://doc.qt.io/qt-5/configure-options.html#opengl-options-for-windows) explain the difference in configuration options. Maybe even though they are considered optional dependencies of Qt Charts, it tries to opportunistically load them?

PatTheMav:
Seems to explicitly require `-DQT_FEATURE_opengl:BOOL=NO` to disable it opportunistically adding it.

The way the logic works is: "If OpenGL support is not disabled, use it". Thus it becomes a hard requirement of QtCharts. We haven't explicitly disabled it so far, but might be required now.

RytoEX:
Are the OpenGL DLLs copied in the obs-studio Qt Charts PR in CMake 3? Does Qt Charts fail to work if the OpenGL DLLs are not in the obs-studio directory?

pkviet:
> Are the OpenGL DLLs copied in the obs-studio Qt Charts PR in CMake 3? Does Qt Charts fail to work if the OpenGL DLLs are not in the obs-studio directory?

Yes currently i'm copying the qt opengl DLLs in the two stats PRs in obs-studio.
(I plan to overhaul them pending mockups from George and ui/ux discussions.)
Qt Charts fails to load without them.
I'm trying to do without the openGL deps following Pat's pointers.
I thus passed -DFEATURE_opengl:BOOL=OFF
But QtGui fails to compile on windows with errors in qwindowswindow.cpp (line 2310):
use of undefined type QOpenGLContext
And another similar error ...

pkviet:
**Update**
With Pat's help I managed to compile without any openGL dependence.

pkviet:
> **Update** With Pat's help I managed to compile without any openGL dependence.

**Update**
With the update from 6.4.3 to 6.5.2, QtMultimedia fails to build when one disables opengl.
So we're back to square one.
Not sure why we would need QtMultimedia btw.

RytoEX:
> > **Update** With Pat's help I managed to compile without any openGL dependence.
> 
> **Update** With the update from 6.4.3 to 6.5.2, QtMultimedia fails to build when one disables opengl. So we're back to square one. Not sure why we would need QtMultimedia btw.

QtMultimedia was to be used for accessibility features to play sound effects on certain actions. It wasn't implemented initially because it required Windows Media features that were not installed by default on Windows N systems, and earlier versions of Qt and FFmpeg would crash if they tried to load the necessary libraries. We pushed for changes in those implementations to safely attempt to load the libraries and just never circled back around to implementing the feature on our end.

RytoEX:
> > **Update** With Pat's help I managed to compile without any openGL dependence.
> 
> **Update** With the update from 6.4.3 to 6.5.2, QtMultimedia fails to build when one disables opengl. So we're back to square one. Not sure why we would need QtMultimedia btw.

The only thing I could find in Qt's Jira was this: https://qt-project.atlassian.net/browse/QTBUG-97485, which says it was fixed in Qt 6.2/6.3. If this is still an issue, we should file a QTBUG for it.

pkviet:
> > > **Update** With Pat's help I managed to compile without any openGL dependence.
> > 
> > 
> > **Update** With the update from 6.4.3 to 6.5.2, QtMultimedia fails to build when one disables opengl. So we're back to square one. Not sure why we would need QtMultimedia btw.
> 
> The only thing I could find in Qt's Jira was this: https://qt-project.atlassian.net/browse/QTBUG-97485, which says it was fixed in Qt 6.2/6.3. If this is still an issue, we should file a QTBUG for it.

Got it working; it requires 3 cmake flags:

>'-DINPUT_opengl=no'
        '-DFEATURE_opengl:BOOL=OFF'   
        '-DQT_NO_OPENGL:BOOL=ON'

RytoEX:
> Got it working; it requires 3 cmake flags:
> 
> > '-DINPUT_opengl=no'
> > '-DFEATURE_opengl:BOOL=OFF'
> > '-DQT_NO_OPENGL:BOOL=ON'

I feel like it should only require the first one. Setting `INPUT_opengl=no` will disable `opengles2`, `opengl-desktop`, `opengl-dynamic`, and `dynamicgl` (`FEATURE_opengles2`, `FEATURE_opengl_desktop`, `FEATURE_opengl_dynamic`), which will also disable `opengl`. If it doesn't, then that seems like a bug in Qt.

deps.qt/qt6.ps1
```diff
@@ -110,6 +111,9 @@ function Configure {
         '-DFEATURE_system_zlib:BOOL=OFF'
         '-DFEATURE_testlib:BOOL=OFF'
         '-DFEATURE_windeployqt:BOOL=OFF'
+        '-DINPUT_opengl=no'
```
RytoEX:
```suggestion
        '-DINPUT_opengl:STRING=no'
```

deps.qt/qt6.ps1
```diff
@@ -110,6 +111,9 @@ function Configure {
...
+        '-DFEATURE_opengl:BOOL=OFF'
+        '-DQT_NO_OPENGL:BOOL=ON'
```
RytoEX: Per my previous comment, I don't think the second two are needed (`'-DFEATURE_opengl:BOOL=OFF'` and `'-DQT_NO_OPENGL:BOOL=ON'`). If they are needed, then sort them appropriately with the rest of the items here.

RytoEX (changes requested):

pkviet:
> > Got it working; it requires 3 cmake flags:
> > > '-DINPUT_opengl=no'
> > > '-DFEATURE_opengl:BOOL=OFF'
> > > '-DQT_NO_OPENGL:BOOL=ON'
> 
> I feel like it should only require the first one. Setting `INPUT_opengl=no` will disable `opengles2`, `opengl-desktop`, `opengl-dynamic`, and `dynamicgl` (`FEATURE_opengles2`, `FEATURE_opengl_desktop`, `FEATURE_opengl_dynamic`), which will also disable `opengl`. If it doesn't, then that seems like a bug in Qt.

nope, keeping only the first flag leads in qtbase>build_x64>config.summary  to:
>  OpenGL:
    Desktop OpenGL ....................... no
    Dynamic OpenGL ....................... yes
    OpenGL ES 2.0 ........................ no
    OpenGL ES 3.0 ........................ no
    OpenGL ES 3.1 ........................ no
    OpenGL ES 3.2 ........................ no

and qtbase fails to compile with 
> qtbase\src\plugins\platforms\windows\qwindowswindow.cpp(2403): error C2653: 'QOpenGLStaticContext': is not a class or namespace name
\qtbase\src\plugins\platforms\windows\qwindowswindow.cpp(2403): error C2065: 'opengl32': undeclared identifier
\qtbase\src\plugins\platforms\windows\qwindowswindow.cpp(2404): error C2027: use of undefined type 'QOpenGLContext'
/qtbase/src/gui/kernel/qsurfaceformat.h(13): note: see declaration of 'QOpenGLContext'

deps.qt/qt6.ps1
pkviet: oddly, with the syntax you advise, it doesn't build

deps.qt/qt6.ps1
pkviet: > > Got it working; it requires 3 cmake flags:
> > > '-DINPUT_opengl=no'
> > > '-DFEATURE_opengl:BOOL=OFF'
> > > '-DQT_NO_OPENGL:BOOL=ON'
> 
> I feel like it should only require the first one. Setting `INPUT_opengl=no` will disable `opengles2`, `opengl-desktop`, `opengl-dynamic`, and `dynamicgl` (`FEATURE_opengles2`, `FEATURE_opengl_desktop`, `FEATURE_opengl_dynamic`), which will also disable `opengl`. If it doesn't, then that seems like a bug in Qt.

nope, keeping only the first flag does lead in qtbase>build_x64>config.summary  to:
>   OpenGL:
    Desktop OpenGL ....................... no
    Dynamic OpenGL ....................... no
    OpenGL ES 2.0 ........................ no
    OpenGL ES 3.0 ........................ no
    OpenGL ES 3.1 ........................ no
    OpenGL ES 3.2 ........................ no

BUT it compiles Qtbase and QtCharts.dll STILL with opengl deps.
And later it fails with qttools>qtdesigner:

 > qttools\src\designer\src\lib\uilib\formbuilder.cpp
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(44): error C3646: 'textureFormat': unknown override specifier
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(44): error C2059: syntax error: '('
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(44): error C2238: unexpected token(s) preceding ';'
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(45): error C2061: syntax error: identifier 'GLenum'
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(54): error C3646: 'defaultFramebufferObject': unknown override specifier
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(54): error C2059: syntax error: '('
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(54): error C2238: unexpected token(s) preceding ';'
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(55): error C3646: 'defaultFramebufferObject': unknown override specifier
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(55): error C2059: syntax error: '('
I:\visual-studio\obs-deps\windows\obs-qt-x64\include\QtOpenGLWidgets/qopenglwidget.h(55): error C2238: unexpected token(s) preceding ';'

RytoEX: > nope, keeping only the first flag does lead in qtbase>build_x64>config.summary to:
> 
> > OpenGL:
> > Desktop OpenGL ....................... no
> > Dynamic OpenGL ....................... no
> > OpenGL ES 2.0 ........................ no
> > OpenGL ES 3.0 ........................ no
> > OpenGL ES 3.1 ........................ no
> > OpenGL ES 3.2 ........................ no
> 
> BUT it compiles Qtbase and QtCharts.dll STILL with opengl deps.

That really seems like a bug to me. We can keep all three to make sure it builds, but this should be reported to Qt.

deps.qt/qt6.ps1
RytoEX: That seems odd, since all of Qt's checks against it are `STREQUAL`.

deps.qt/qt6.ps1
pkviet: i retract what i said; '-DINPUT_opengl:STRING=no' is enough to build.
I had to delete the windows folder in obs-deps; cleaning up windows_build_temp was not enough ...

deps.qt/qt6.ps1
pkviet: i had to clean up windows folder too because a later rebuild pulls libs from there. After that it worked

PR #193, opened 2023-06-21
deps.ffmpeg: Add libdatachannel Simulcast support

Sean-Der:
This allows broadcasters to send the same scene but with multiple quality levels.

Landed upstream in f39a85a39fca38e294b44b8c23c69ab32102bea6

### Motivation and Context
These are the benefits over doing server side transcodes.

* Better quality - Quality is lost server side when doing a decode/re-encode to generate renditions
* Lower latency - Video doesn't need to be processed server side, can be passed directly to viewers
* Greater control - Broadcasters can pick what quality levels they want. Custom framerates/resolutions etc...
* Easier to run servers - Standing up 'broadcast servers' will be easier. Don't need to worry about transcoding etc..

### How Has This Been Tested?
Tested against libdatachannel only. I will be opening PR against OBS next to use this feature.

### Types of changes

### Checklist:


Sean-Der:
@pkviet done!

PR #195, opened 2023-06-23
deps.windows: Add VPL

thyintel:
### Description

Adds vpl v2023.3.0 (latest) headers for obs-qsv11.
Linked with PR: https://github.com/obsproject/obs-studio/pull/9142

### Motivation and Context

Updates Intel MSDK to Intel VPL API to support future Intel hardware. Removes the MSDK files from the obs-studio repo with deprecated MSDK functions and links the VPL headers to obs-qsv11.

### How Has This Been Tested?

Testing is described in the linked obs-studio PR 

### Types of changes

### Checklist:

- [x ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x ] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x ] My code is not on the master branch.
- [x ] The code has been tested.
- [x ] All commit messages are properly formatted and commits squashed where appropriate.
- [x ] I have included updates to all appropriate documentation.

### DIFF
--- a/deps.windows/70-vpl.ps1
+++ b/deps.windows/70-vpl.ps1
@@ -1,0 +1,95 @@
+param(
+    [string] $Name = 'vpl',
+    [string] $Version = 'v2023.3.0',
+    [string] $Uri = 'https://github.com/oneapi-src/oneVPL.git',
+    [string] $Hash = 'e12ace9761bb52786409e830f619916b86e87fc5',
+    [switch] $ForceStatic = $true
+)
+
+function Setup {
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
+}
+
+function Clean {
+    Set-Location $Path
+
+    if ( Test-Path "build_${Target}" ) {
+        Log-Information "Clean build directory (${Target})"
+        Remove-Item -Path "build_${Target}" -Recurse -Force
+    }
+}
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    if ( $ForceStatic -and $script:Shared ) {
+        $Shared = $false
+    } else {
+        $Shared = $script:Shared.isPresent
+    }
+
+    $OnOff = @('OFF', 'ON')
+    $Options = @(
+        $CmakeOptions
+        '-DBUILD_DISPATCHER_ONLY:BOOL=ON'
+        '-DBUILD_EXAMPLES:BOOL=OFF'
+        '-DBUILD_DISPATCHER_ONEVPL_EXPERIMENTAL:BOOL=OFF'
+        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$Shared])"
+    )
+
+    Invoke-External cmake -S . -B "build_${Target}" @Options
+}
+
+function Build {
+    Log-Information "Build (${Target})"
+    Set-Location $Path
+
+    $Options = @(
+        '--build', "build_${Target}"
+        '--config', $Configuration
+    )
+
+    if ( $VerbosePreference -eq 'Continue' ) {
+        $Options += '--verbose'
+    }
+
+    Invoke-External cmake @Options
+}
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location $Path
+
+    $Params = @{
+        ErrorAction = "SilentlyContinue"
+        Path = @(
+            "$($ConfigData.OutputPath)/lib"
+            "$($ConfigData.OutputPath)/include"
+        )
+        ItemType = "Directory"
+        Force = $true
+    }
+	
+    New-Item @Params *> $null
+
+    $Items = @(
+        @{
+            Path = "api/vpl"
+            Destination = "$($ConfigData.OutputPath)/include"
+            Recurse = $true
+            ErrorAction = 'SilentlyContinue'
+        }
+        @{
+            Path = "build_${Target}/$Configuration/vpl.lib"
+            Destination = "$($ConfigData.OutputPath)/lib"
+            ErrorAction = 'SilentlyContinue'
+        }
+    )
+	
+    $Items | ForEach-Object {
+        $Item = $_
+        Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
+        Copy-Item @Item
+    }
+}
--- a/licenses/vpl/LICENSE.txt
+++ b/licenses/vpl/LICENSE.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Intel Corporation
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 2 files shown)


RytoEX (changes requested):

RytoEX (changes requested):

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $Options = @(
+        $CmakeOptions
+        ''
+    )
```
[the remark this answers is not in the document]
thyintel: Added the Cmake flags, should be the bare minumim that is building for vpl.

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,63 @@
+param(
+    [string] $Name = 'vpl',
+    [string] $Version = 'v2023.3.0',
+    [string] $Uri = 'https://github.com/oneapi-src/oneVPL.git',
+    [string] $Hash = '6e9f56aacbcb3b4ad1800cba091aaf9ec32135f9'
```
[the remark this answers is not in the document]
thyintel: Switched to the more updated commit!

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $Options = @(
+        $CmakeOptions
+        '-DBUILD_DISPATCHER_ONLY:BOOL=ON'
+		'-DBUILD_EXAMPLES:BOOL=OFF'
+		'-DBUILD_DISPATCHER_ONEVPL_EXPERIMENTAL_ONLY:BOOL=OFF'
+		'-DBUILD_SHARED_LIBS:BOOL=OFF'
```
RytoEX, line 30: Properly align these with spaces. See other Windows dependency scripts for examples.

RytoEX, line 30: @PatTheMav should this use the $OnOff syntax to be controllable rather than hard-coded?

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+			ErrorAction = 'SilentlyContinue'
+		}
+	)
+	
+	$Items | ForEach-Object {
+		$Item = $_
+		Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
+		Copy-Item @Item
+	}
+}
```
RytoEX, line 87: Use spaces, not tabs, to align.

RytoEX (changes requested):
Mostly nits for now.

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+            ErrorAction = 'SilentlyContinue'
+        }
+    )
+	
+    $Items | ForEach-Object {
+        $Item = $_
+        Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
+        Copy-Item @Item
+    }
+}
```
RytoEX, line 87: Always end a file with a blank newline.

RytoEX (changes requested):
Commit message and PR title nit:
* vpl -> VPL

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $Options = @(
+        $CmakeOptions
+        '-DBUILD_DISPATCHER_ONLY:BOOL=ON'
+        '-DBUILD_EXAMPLES:BOOL=OFF'
+        '-DBUILD_DISPATCHER_ONEVPL_EXPERIMENTAL_ONLY:BOOL=OFF'
```
RytoEX, line 29: The flag is actually `BUILD_DISPATCHER_ONEVPL_EXPERIMENTAL`.

RytoEX (changes requested):
This will also need the repo's LICENSE file copied to an appropriate sub-directory in `licenses/`.

RytoEX:
As an aside, this PR does more than just add headers, since it also appears to build `vpl.lib`. If `vpl.lib` is a required component, then please update the PR title and commit message so that this reflects that. Something like simply "Add VPL" is probably sufficient to avoid being misleading.

RytoEX:
> As an aside, this PR does more than just add headers, since it also appears to build `vpl.lib`. If `vpl.lib` is a required component, then please update the PR title and commit message so that this reflects that. Something like simply "Add VPL" is probably sufficient to avoid being misleading.

While the PR title was updated, the commit message still uses the old wording.

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+        @{
+            Path = "build_${Target}/$Configuration/vpl.lib"
+            Destination = "$($ConfigData.OutputPath)/lib"
+            ErrorAction = 'SilentlyContinue'
+        }
+    )
+	
+    $Items | ForEach-Object {
+        $Item = $_
+        Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
```
PatTheMav, line 84: I'm confused by what `$Item.Path -join ", "` is supposed to achieve, given that it's always a relative path string and not an array (given that this is code wrapped in a `ForEach-Object` block).

thyintel: I was using the other deps.windows files as a reference and this was the syntax that they used. If you have a better/simpler way to execute this function, I am open to any suggestions!

PatTheMav: Checked the sources - it is indeed unnecessary in this specific case, but we also have cases where the `Path` element of the hashtable is indeed an array:

```
    $Params = @{
        ErrorAction = "SilentlyContinue"
        Path = @(
            "$($ConfigData.OutputPath)/bin"
            "$($ConfigData.OutputPath)/lib"
        )
        ItemType = "Directory"
        Force = $true
    }
```

It's fine to use this as a generic pattern then.

RytoEX (changes requested):
My last nit is mostly that I don't think we need to mention "obs-qsv11". It's possible that the module will not always be named that and that the dependency may be used by other components. Just "Add VPL" should be sufficient.

@PatTheMav ? @Kurufu ? Any final concerns?

RytoEX (approved):
Looks good to me overall. Would still like a follow-up to my one open question to @PatTheMav about hard-coding `BUILD_SHARED_LIBS` to OFF.

deps.windows/70-vpl.ps1
PatTheMav: Should use the existing pattern instead of hardcoding. We might revise this in the future as we have a growing number of dependencies with different usage requirements (some can _only_ be used as shared libraries, some can _only_ be used as static ones, some are _preferred_ to be used as shared libraries, and so on), but that's the requirement for now.

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
+param(
+    [string] $Name = 'vpl',
+    [string] $Version = 'v2023.3.0',
+    [string] $Uri = 'https://github.com/oneapi-src/oneVPL.git',
+    [string] $Hash = 'e12ace9761bb52786409e830f619916b86e87fc5'
```
PatTheMav, line 5: Add `[switch] $ForceStatic = $true` here.

deps.windows/70-vpl.ps1
```diff
@@ -0,0 +1,87 @@
...
+        Log-Information "Clean build directory (${Target})"
+        Remove-Item -Path "build_${Target}" -Recurse -Force
+    }
+}
+
+function Configure {
+    Log-Information "Configure (${Target})"
+    Set-Location $Path
+
+    $Options = @(
```
PatTheMav, line 25: Should probably use the following instead:

```
    if ( $ForceStatic -and $script:Shared ) {
        $Shared = $false
    } else {
        $Shared = $script:Shared.isPresent
    }    

    $OnOff = @('OFF', 'ON')
    $Options = @(
        $CmakeOptions
[...]
        "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$Shared])"
    )
```

We probably might simplify this across the board later, but for now that's the common pattern.

deps.windows/70-vpl.ps1
RytoEX: I've added this via force push.

deps.windows/70-vpl.ps1
RytoEX: I've added this via force push.

deps.windows/70-vpl.ps1
RytoEX: I've addressed this via force push. Resolving this thread.

PR #196, opened 2023-06-30
Remove non-essential Windows x86 deps builds

PatTheMav:
### Description
Removes x86 slices of most Windows dependencies, with the exception of Detours and Vulkan, which are both needed for capture plugins.

### Motivation and Context
Deprecate all traces of standalone x86 OBS Studio versions.

### How Has This Been Tested?
Tested on Windows 11 VM.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -150,12 +150,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        target: [x64, x86]
+        target: [x64]
         include:
           - target: x64
-            config: Release
-            type: static
-          - target: x86
             config: Release
             type: static
     defaults:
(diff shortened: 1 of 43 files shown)


RytoEX (changes requested):
Do the first two commits individually work in the absence of the third commit? It looks like they should, but I wanted to confirm.

Commit message nit:
* detours -> Detours

PatTheMav:
> Do the first two commits individually work in the absence of the third commit? It looks like they should, but I wanted to confirm.
> 
> Commit message nit:
> 
> * detours -> Detours

They should, yeah.

RytoEX (approved):
Seems fine. The one thing to watch out for is that it will become impossible to do full 32-bit builds in legacy CMake, so it may be helpful (though perhaps not required) to have instructions for a slim 32-bit build with just the win-capture and win-dshow components.

RytoEX:
Per off-thread discussion, we're going to hold off on merging this change because removing the x86 Windows deps would make doing complete Windows legacy CMake builds of obs-studio a bit burdensome. We'll remove the x86 Windows deps once we remove the legacy CMake.

RytoEX:
Can this be merged before #190?

RytoEX:
Let's cut down CI's workload then.

PR #198, opened 2023-07-17
CI: Update Ubuntu runners to 22.04

RytoEX:
### Description

CI: Update Ubuntu runners to 22.04

This updates the Ubuntu runners used for the Linux/Ubuntu FFmpeg test job and also the Windows FFmpeg cross-compilation jobs. Incidentally, this might also fix the current CI failures caused by nasm.us being down, since 22.04 has a newer version of nasm available (2.15.05 on Ubuntu 22.04 versus 2.14.02 on Ubuntu 20.04).

### Motivation and Context

Want the runners to use the minimum version of Ubuntu we support. Don't want CI to fail.

### How Has This Been Tested?

It has not been tested.

### Checklist:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -72,17 +72,17 @@ jobs:
             type: 'static'
             revision: 9
           - target: linux-x86_64
-            os: 'ubuntu-20.04'
+            os: 'ubuntu-22.04'
             config: 'Release'
             type: 'static'
             revision: 9
           - target: windows-x64
-            os: 'ubuntu-20.04'
+            os: 'ubuntu-22.04'
             config: 'Release'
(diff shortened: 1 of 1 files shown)


RytoEX:
For added context, obs-deps CI is failing because `nasm_check` looks for 2.15.01, but Ubuntu 20.04 only has 2.14.02 available via apt. As a result, we attempt to download and build 2.15.01 manually from nasm.us. Since nasm.us is down/unreachable, that fallback fails, and the CI FFmpeg Ubuntu and Windows jobs fail. Ubuntu 22.04 has nasm 2.15.05 available via apt, so the fallback is not used.

PatTheMav (approved):

PR #200, opened 2023-07-18
Update Windows build script to use explicit package names

PatTheMav:
### Description
Updates the Windows build script to explicitly identify dependency "package names" (instead of inferring them implicitly).

### Motivation and Context
Allows checking for package name globally and adopt different code paths for each package.

Prerequisite for https://github.com/obsproject/obs-deps/pull/186 as caches generated by nightly runs are incompatible with that codebase otherwise.

This change adds new parameter `PackageName` to the build script which selects the kind of dependency package to build:

* `dependencies` (the default)
* `qt`
* `ffmpeg` (added later in https://github.com/obsproject/obs-deps/pull/186)

When the parameter is omitted, default Windows dependencies are built. If Qt is to be built `-PackageName qt` needs to be provided instead.

### How Has This Been Tested?
Needs to be tested on CI and later with #186 after merging and rebasing.

### Types of changes

### Checklist:

### DIFF
--- a/Build-Dependencies.ps1
+++ b/Build-Dependencies.ps1
@@ -2,6 +2,8 @@
 param(
     [ValidateSet('Debug', 'RelWithDebInfo', 'Release', 'MinSizeRel')]
     [string] $Configuration = 'Release',
+    [ValidateSet('dependencies', 'qt')]
+    [string] $PackageName = 'dependencies',
     [string[]] $Dependencies,
     [ValidateSet('x86', 'x64')]
     [string] $Target,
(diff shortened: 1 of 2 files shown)


Build-Dependencies.ps1
```diff
@@ -114,34 +116,25 @@ function Run-Stages {
...
-    )
-
-    Remove-Item -ErrorAction 'SilentlyContinue' -Path $Items -Force -Recurse
+
+    switch ( $PackageName ) {
+        dependencies {
+            Get-ChildItem ./bin/*.lib | Move-Item -Destination ./lib
+            Get-ChildItem ./bin/* -Exclude 'libcurl.dll','lua51.*,','Lib','swig.exe' | Remove-Item
+            Get-ChildItem ./cmake/pcre2*,./lib/pcre2* | Remove-Item
+            Remove-Item -Recurse ./lib/pkgconfig
```
RytoEX, line 128: Are we making an intentional decision to remove all pkgconfig files?

Build-Dependencies.ps1
```diff
@@ -114,34 +116,25 @@ function Run-Stages {
...
+            Remove-Item -Recurse ./man
+            Get-ChildItem ./share/* -Exclude 'nlohmann?json*' | Remove-Item -Recurse
```
RytoEX, line 130: Is the single character wildcard intentional?

RytoEX (changes requested):
What would the new build invocations look like with these changes?

Build-Dependencies.ps1
```diff
@@ -114,34 +116,25 @@ function Run-Stages {
...
-        "./share/bison"
-        "./share/doc/pcre2"
-    )
-
-    Remove-Item -ErrorAction 'SilentlyContinue' -Path $Items -Force -Recurse
+
+    switch ( $PackageName ) {
+        dependencies {
+            Get-ChildItem ./bin/*.lib | Move-Item -Destination ./lib
+            Get-ChildItem ./bin/* -Exclude 'libcurl.dll','lua51.*,','Lib','swig.exe' | Remove-Item
```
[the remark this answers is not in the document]
PatTheMav: That is correct for the main dependencies, as only cURL is shipped as a DLL and swig with its `Lib` directory is the only non-library file needed in there.

Everything else is a static library that was moved into `lib` with the command before that.

Build-Dependencies.ps1
PatTheMav: Correct, pkg-config is not used outside of Linux/FreeBSD in our case.

Build-Dependencies.ps1
PatTheMav: Correct, nothing apart from the nlohmann files is needed in `shared` (Swig's library files on Windows are always put in a directory called `Libs` adjacent to the binary).

PatTheMav: ("nlohmann files" consists of CMake and pkg-config files, I didn't bother to break that down further to also remove the pkg-config directory here).

Build-Dependencies.ps1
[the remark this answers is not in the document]
PatTheMav: Ah right, that was introduced after this was authored - will do a pass over those.

Build-Dependencies.ps1
RytoEX: I meant the specific format of `nlohmann?json*` versus `nlohmann_json*`. Are there files that do not match the latter? I could only find files that match the latter more explicit pattern in the current deps package.

PatTheMav: Cannot really say after all these months - I _guess_ this might have been because of a mismatch between the dependency name "nlohmann-json" and the generated file names based on "nlohmann_json", but you're right that the generated files seem to follow the latter pattern only.

Build-Dependencies.ps1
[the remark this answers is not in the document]
PatTheMav: Found the main reason - this PR was backported from the FFmpeg update, where libdatachannel is moved to the FFmpeg part (where the macOS variant is placed as well) and as such would never be built in this slice.

I'll add it for now and remove it again with the FFmpeg update.

RytoEX (commented):
Probably fine. I'll try to do some quick local testing before giving full approval for next merge window.

RytoEX (approved):
Worked fine locally. Seems fine.

PR #203, opened 2023-07-24
Add uthash

tytan652:
### Description

Add uthash to obs-deps.

Yes, they should be installed in the include directory without any subdirectory to align with Linux distros.

### Motivation and Context

Reduce in-tree vendored deps in OBS Studio repo.

### How Has This Been Tested?

CI builds artifacts on my fork PR, they contain headers and license files.

### Types of changes

### Checklist:


derrod:
Is there a specific reason to include the other headers? We only need `uthash.h` and there should be no interdependence between them.

tytan652:
I just did it like a packager should do it since all the header are part of uthash.

deps.windows/60-uthash.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location "$Path"
+
+    New-Item -Path "$($ConfigData.OutputPath)/include" -ItemType Directory -Force *> $null
+
+    $Items = @(
+        @{
+            Path = "src/utarray.h"
```
PatTheMav, line 20: You should be able to specify an array here (`"src/utarray.h", "src/uthash.h", [..]`), PowerShell usually accepts that as input parameter for file operations.

tytan652: No log is fine, or do you want some kind of logging ?

I ask before I resolve this and squash my commits.

PatTheMav: Why did you remove the rest of the code? That code was already compatible with providing an array (hence why it did the whole thing about joining the string). The only change necessary was to have a single item with the files provided as an array. :D

tytan652: You say to use an array which usually make the whole for loop useless (and so the log per file).

tytan652: So you want to keep the loop ?

PatTheMav (approved):
Should be good. I also agree with copying all related files to make it a "default" distribution.

The compiler will pull in files as needed.

Conan-Kudo (approved):

RytoEX (approved):
Seems fine at a glance. Will merge when we get to a merge window.

PR #210, opened 2023-09-07
CI: Enable qrcodegencpp Debug build

admshao:
### Description
Enables Debug builds of qrcodegencpp dependency

### Motivation and Context
Enables Debug builds without errors on obs-studio

### How Has This Been Tested?
Built the dependency with the following command and manually copied the generated files to obs-studio/.deps/.../ folder.
```
$Params = @{
>>     Target = 'x64'
>>     Configuration = 'Debug'
>>     Dependencies = 'qrcodegencpp'
>>   }
.\Build-Dependencies.ps1 @Params
```
Checked the WebSocket was enabled once again on Debug builds.

### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/60-qrcodegencpp.ps1
+++ b/deps.windows/60-qrcodegencpp.ps1
@@ -52,6 +52,7 @@ function Configure {
     $OnOff = @('OFF', 'ON')
     $Options = @(
         $CmakeOptions
+        '-DCMAKE_DEBUG_POSTFIX:STRING=d'
         "-DBUILD_SHARED_LIBS:BOOL=$($OnOff[$script:Shared.isPresent])"
     )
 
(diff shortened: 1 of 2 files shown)


RytoEX:
This is not quite complete, as I discovered yesterday. The Qrcodegen project builds Debug libraries with the same names as Release libraries, so they will clobber each other. I fixed it locally yesterday, but got bogged down in other things and didn't push the work. I can push corrections to the build script independently (or force-push them to this branch), as this work is otherwise correct.

The historically used commit prefix for this file is "CI: ".

deps.windows/60-qrcodegencpp.ps1
```diff
@@ -108,4 +108,17 @@ function Fixup {
             Remove-Item -Recurse -Force $Item.FullPath
         }
     }
+    $Items = @(
```
PatTheMav, line 111: As mentioned off-thread, this should either be handled by setting `CMAKE_DEBUG_POSTFIX` (to enforce the `d` suffix) or if the file is renamed manually, the generated `qrcodegencpp-targets-debug.cmake` file needs to be patched with the new filename (as otherwise CMake will pick up the non-debug variant for debug again).

RytoEX: I've addressed this in a force push to use `CMAKE_DEBUG_POSTFIX`, which is also what I had done locally and did not have time to push up anywhere.

RytoEX:
I force pushed a minor addition to fixup to also remove `qrcodegend.lib`, which is now created when building in Debug due to the setting CMAKE_DEBUG_POSTFIX. PR is otherwise the same.

RytoEX (approved):
Should be fine. I've checked these changes locally.

PR #211, opened 2023-09-09
deps.qt: Backport option to disable native messagebox dialog

gxalpha:
### Description

Backports qt/qtbase@29b2506e which adds the option to disable native messageboxes on macOS.
Unfortunately, the native messageboxes were introduced in 6.5 but the option only exists in 6.6, which is why this backport is needed. To be able to differentiate the patched 6.5 from a normal one, the patch was amended to add a `#define OBS_QT_NON_NATIVE_MESSAGEBOX_DIALOG` that can then be checked against in OBS source code to check for the availability of the option.

### Motivation and Context

Needed to make obsproject/obs-studio#9579 actually do something.

### How Has This Been Tested?

Tested with a [patched release of obs-deps](https://github.com/gxalpha/obs-deps/releases/tag/2023-09-09) and the changes in obsproject/obs-studio#9579.

### Types of changes

### Checklist:


gxalpha:
This is not actually needed if we use some small hacks on obs-studio side.

PR #212, opened 2023-09-11
deps.qt: Fix library naming for macOS debug builds

PatTheMav:
### Description
Fixes the output file names and library IDs of Qt debug builds on macOS by forcing the use of unversioned `SONAME`s.

### Motivation and Context
The Xcode bundle generation requires libraries generated by CMake to not use versioned SONAMEs (as Linux does), which Qt does by default.

Disabling this behaviour generates unversioned dynamic libraries and makes them compatible with `obs-studio`'s build system.

With this change, a local Qt debug build can be copied over downloaded `obs-deps` and will be picked up whenever the project itself is built in Debug configuration.

### How Has This Been Tested?
Tested with Debug builds on macOS 13.5.2 and current Xcode.

### Types of changes

### Checklist:

### DIFF
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -98,7 +98,8 @@ config() {
     fi
   }
 
-  if (( shared_libs)) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON -DQT_FEATURE_force_debug_info:BOOL=ON)
+  if (( shared_libs )) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON -DQT_FEATURE_force_debug_info:BOOL=ON)
+  if (( shared_libs )) && [[ ${config} == Debug ]] common_cmake_flags+=(-DCMAKE_PLATFORM_NO_VERSIONED_SONAME:BOOL=ON)
 
   args=(
     ${common_cmake_flags}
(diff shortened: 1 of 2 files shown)


deps.qt/qt6.zsh
```diff
@@ -98,7 +98,8 @@ config() {
     fi
   }
 
-  if (( shared_libs)) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON -DQT_FEATURE_force_debug_info:BOOL=ON)
+  if (( shared_libs )) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON -DQT_FEATURE_force_debug_info:BOOL=ON)
+  if (( shared_libs)) && [[ ${config} == Debug ]] common_cmake_flags+=(-DCMAKE_PLATFORM_NO_VERSIONED_SONAME:BOOL=ON)
```
RytoEX, line 102:
```suggestion
  if (( shared_libs )) && [[ ${config} == Debug ]] common_cmake_flags+=(-DCMAKE_PLATFORM_NO_VERSIONED_SONAME:BOOL=ON)
```

deps.qt/qt6.zsh
```diff
@@ -200,6 +201,7 @@ qt_add_submodules() {
   }
 
   if (( shared_libs)) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON)
+  if (( shared_libs)) && [[ ${config} == Debug ]] common_cmake_flags+=(-DCMAKE_PLATFORM_NO_VERSIONED_SONAME:BOOL=ON)
```
RytoEX, line 204:
```suggestion
  if (( shared_libs )) && [[ ${config} == Debug ]] common_cmake_flags+=(-DCMAKE_PLATFORM_NO_VERSIONED_SONAME:BOOL=ON)
```

deps.qt/qt6.zsh
```diff
@@ -200,6 +201,7 @@ qt_add_submodules() {
   }
 
   if (( shared_libs)) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON)
```
RytoEX, line 203:
```suggestion
  if (( shared_libs )) && [[ ${config} == Release ]] common_cmake_flags+=(-DFEATURE_separate_debug_info:BOOL=ON)
```

RytoEX (changes requested):
Just whitespace nits on the `if` statements. Looks like we use `if (( condition ))` mostly through, except these few places where we do `if (( condition))`, which was probably just a typo that slipped through previously.

RytoEX (approved):
Seems fine. Will merge once we get back into a merge window for deps.

PR #213, opened 2023-09-11
deps.ffmpeg: Fix local universal architecture builds of aom for macOS

PatTheMav:
### Description
Fixes universal builds of `aom` on macOS by extending the functionality available for non-CMake-based projects to also override CMake variables with architecture-specific identifiers.

### Motivation and Context
`aom` requires architecture-specific toolchain files to be built, but universal builds usually only require passing both modern Apple architectures to CMake to have the compiler create universal binaries by default.

Because of architecture-specific optimisations this is not possible for `aom`, so for local builds the architectures need to be built separately and then combined for the final installation.

### How Has This Been Tested?
Tested locally by building `aom` as static and shared libraries and checking the available architectures in the installed library files.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/40-aom.zsh
+++ b/deps.ffmpeg/40-aom.zsh
@@ -50,6 +50,11 @@ patch() {
 config() {
   autoload -Uz mkcd progress
 
+  if [[ ${target} == macos-universal ]] {
+      autoload -Uz universal_config && universal_config
+      return
+  }
+
   local _onoff=(OFF ON)
 
   args=(
@@ -76,6 +81,11 @@ config() {
 
 build() {
   autoload -Uz mkcd progress
+
+  if [[ ${target} == macos-universal ]] {
+      autoload -Uz universal_build && universal_build
+      return
+  }
 
   log_info "Build (%F{3}${target}%f)"
 
@@ -104,6 +114,13 @@ install() {
   if (( _loglevel > 1 )) args+=(--verbose)
 
   cd ${dir}
+
+  if [[ ${target} == macos-universal ]] {
+    pushd build_universal
+    sed -i -E -e 's/build_x86_64/build_universal/g' cmake_install.cmake
+    popd
+  }
+
   progress cmake ${args}
 }
 
@@ -148,7 +165,6 @@ fixup() {
         rm -rf ${target_config[output_dir]}/bin/libaom*.dll(N)
       }
       ;;
-    }
   }
 
   if (( #strip_files )) && [[ ${config} == (Release|MinSizeRel) ]] ${strip_tool} -x ${strip_files}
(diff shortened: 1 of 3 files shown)


deps.ffmpeg/40-aom.zsh
```diff
@@ -148,7 +165,6 @@ fixup() {
         rm -rf ${target_config[output_dir]}/bin/libaom*.dll(N)
       }
       ;;
-    }
```
RytoEX, line 151: Doesn't this match the opening brace for the `case` statement at line 118/135?

deps.ffmpeg/40-aom.zsh
```diff
@@ -104,6 +114,13 @@ install() {
   if (( _loglevel > 1 )) args+=(--verbose)
 
   cd ${dir}
+
+  if [[ ${target} == macos-universal ]] {
+    pushd build_universal
+    sed -i -E -e 's/build_x86_64/build_universal/g' cmake_install.cmake
```
RytoEX, line 120: What's the rationale on this?

utils.zsh/functions/universal_build
```diff
@@ -3,6 +3,7 @@ autoload -Uz log_info log_status
 local arch
 for arch (x86_64 arm64) {
   (
+    typeset -g -A target_config=(${(kv)target_config//x86_64;arm64/${arch}})
```
RytoEX, line 6: In `utils.zsh/functions/universal_config`, this near the end of the list. Should they be aligned for consistency?

RytoEX (changes requested):

deps.ffmpeg/40-aom.zsh
PatTheMav: Because `aom` cannot be built as a universal library, the installed files from the `x86_64` and `arm64` build directories are combined into a universal build after the fact.

The cmake install script inherited from the `x86_64` will still retain references to that architecture however, so it would install files from there instead of the manually created `build_universal` directory.

This patch ensures that the install script will copy the "fat" binaries created by the mixup.

RytoEX: Got it. A little unfortunate, but oh well.

deps.ffmpeg/40-aom.zsh
PatTheMav: Nope, that's what the next one does. Currently the script actually skips the stripping step below, removing this brace actually fixes it.

RytoEX: I checked in a text editor, and now I understand. Thanks.

RytoEX (approved):
Seems fine at a glance. Will merge when we get to an appropriate merge window.

PR #216, opened 2023-09-23
deps.ffmpeg: Fix libdatachannel packaging for macOS

PatTheMav:
### Description
Fix malformed `SONAME` and library name in CMake package file generated for `libdatachannel`.

Fixes current issue with `obs-webrtc` failing to load in OBS Studio on macOS because the dependency library cannot be resolved by the dynamic library loader.

### Motivation and Context
Libraries should not be symlinked or versioned on macOS, so we need to fix not only the generated binaries but also the CMake package files to reflect this.

### How Has This Been Tested?
Built `libdatachannel` locally with the fixes applied and replaced corresponding contents in downloaded `obs-deps`, then built and ran OBS Studio to check if `obs-webrtc` is correctly loaded.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/70-libdatachannel.zsh
+++ b/deps.ffmpeg/70-libdatachannel.zsh
@@ -98,6 +98,10 @@ fixup() {
 
         if [[ ${config} == Release ]] dsymutil ${dylib_files}
 
+        for dylib_file (${dylib_files}) {
+          sed -i '' -E -e "s/${dylib_file:t}/libdatachannel.dylib/g" ${target_config[output_dir]}/lib/cmake/LibDataChannel/LibDataChannelTargets-${(L)config}.cmake
+        }
+
         strip_tool=strip
         strip_files=(${dylib_files})
       } else {


RytoEX (approved):
Seems fine. Is this something already fixed upstream, or something that *should* be fixed upstream?

RytoEX:
My question was more of a "if there's a canonical best practice for packaging libraries on macOS that dependencies should be following, should we be upstreaming such fixes for their benefit, but also to reduce the amount of fixups that we need to perform?"

This is fine to merge as-is without any work upstream. I was just exploring some forward thinking.

Conan-Kudo:
> In theory we probably would need to upstream patches to all dependencies or probably even CMake itself, because they all package macOS libraries as if it were Linux and just replace .so. with .dylib which is conceptually wrong.

Versioned libraries are absolutely fine on macOS, and they exist in the base operating system too. Versioned libraries follow the same convention as on Windows for DLLs: `<name>-<sover>.<ext>`. I actually don't understand why you needed to do this in the first place.

PatTheMav:
If you check the macOS SDK library folder and the system library folder you will notice that the majority of libraries has _no_ version information in its filename. The simple reason is that macOS has no need for them, because version and version-compatibility information is contained in the MACH headers and the library loader does not use the file name for compatibility checks.

Conan-Kudo:
That is true, though you still need to do something if multiple versions of a library are used system-wide. It's far less of a problem when the libraries are part of an app bundle, but things like Homebrew and whatever don't get to benefit from that.

PatTheMav:
OBS Studio is an app and is shipped in an app bundle. 

App bundles do not contain versioned libraries.

Case in point - Affinity Designer's `Frameworks` directory:

```
libcocoaui.framework  liblibbook.dylib      liblibde265.dylib      liblibexpat.dylib     liblibhdphoto.dylib   liblibidmlimport.dylib  liblibjpeg2k.dylib   libliblittlecms.dylib  liblibopenexr.dylib  liblibplugins.dylib  liblibrastertools.dylib  liblibscripting.dylib  liblibtiff.dylib  liblibzstd.dylib
liblibacs.dylib       liblibcfitsio.dylib   liblibdwgimport.dylib  liblibgeometry.dylib  liblibheif.dylib      liblibipc.dylib         liblibjsoncpp.dylib  liblibmetadata.dylib   liblibp2p.dylib      liblibpng.dylib      liblibraw.dylib          liblibstemmer.dylib    liblibwebp.dylib  libpdfimport.dylib
liblibaffinity.dylib  liblibcommands.dylib  liblibeps.dylib        liblibgif.dylib       liblibhunspell.dylib  liblibjbig2dec.dylib    liblibkernel.dylib   liblibnetwork.dylib    liblibpersona.dylib  liblibpsd.dylib      liblibregex.dylib        liblibstory.dylib      liblibxml.dylib   libpdflib.dylib
liblibbmp.dylib       liblibcurl.dylib      liblibepub.dylib       liblibgl.dylib        liblibicu.dylib       liblibjpeg.dylib        libliblf.dylib       liblibocio.dylib       liblibpk.dylib       liblibraster.dylib   liblibrenderer.dylib     liblibtga.dylib        liblibzlib.dylib  libxmp.dylib
```

Here's Cyberduck's:

```
JavaNativeFoundation.framework  Sparkle.framework  libcore.dylib  libjdns_sd.dylib  libjnidispatch.dylib  librococoa.dylib
```

It is _not_ canonical to have versioned dynamic libraries file names in macOS.

Conan-Kudo:
You're right, and I think I was agreeing with you in my previous comment? But it's not wrong for CMake to default to versioned libraries if the defaults are oriented around system-wide installs. Things like homebrew still need those versioned filenames.

PatTheMav:
That's because the vast majority of Homebrew formulas are command line tools primarily written for Linux and as such follow its customs.

The underlying issue is that while macOS can make a distinction between CLI tools and applications (and their different library naming conventions/requirements), there is no such distinction on Linux. And CMake doesn't make this distinction either.

The only way to turn off the naming when creating libraries _specifically for bundling_ is to set the `NO_SONAME` target property (which does have side-effects) or do the patching we do on `obs-deps`. It would be great if one could tell CMake that directly so it also knows the intended use of a built library, but alas it doesn't exist.

Conan-Kudo:
Yeah, I battled this issue with Lugaru, ultimately I decided "meh" about it... But it would be nice if CMake had an "app bundle" mode.

PR #217, opened 2023-09-24
deps.qt: Fix C++17 compile issues with AppleClang 15

PatTheMav:
### Description
Fixes compilation errors with AppleClang 15.

### Motivation and Context
AppleClang 15 implements the C++17 feature "polymorphic memory resources", but the linker guards this functionality behind the macOS 14 deployment target.

The guards implemented in Qt code cannot check for the deployment target, so to keep the C++17 standard, we need to patch out all uses of memory targets in code.

### How Has This Been Tested?
Successfully compiled Qt 6.5.2 on macOS 13.6 with AppleClang 15.

### Types of changes

### Checklist:


RytoEX:
As mentioned previously, I would like to see a QTBUG reporting this or an upstream commit that already solves this before adopting more patches.

PatTheMav:
Apparently they fixed this in 6.5.3 if I read this right: https://bugreports.qt.io/browse/QTBUG-117484

RytoEX:
Was merged as https://github.com/qt/qtbase/commit/c70145c6bf54cda9723cccdc7ca07f8d8fd321cd

PR #223, opened 2023-10-31
CI: Fix issues introduced by CMake version changes when using build caches

PatTheMav:
### Description
Adds the CMake version used on a runner to the generated hash used for build caches and adds VPL and QR-Code-Generator to scheduled builds.

### Motivation and Context
CMake's installation scripts will automatically remove old CMake package installations if the `<project>-targets.cmake` file to be installed differs from the one found at the installation destination.

This process will also remove any configuration-related target files found in that directory. This currently leads to the restored build directory from the GitHub Action cache to remove the files installed by the step to create the Debug configuration variant.

Forcing the cache restoration to miss if the CMake version used for the cache differs from the one currently installed on runners ensures that both steps will use the same underlying CMake version.

### How Has This Been Tested?
Will require CMake changes on a runner to show its actual effectiveness, but it was confirmed that using the same CMake version for a debug configuration run followed by a release run will not remove the target files.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/build-deps/action.yml
+++ b/.github/actions/build-deps/action.yml
@@ -37,6 +37,8 @@ runs:
             ;;
         esac
 
+        deps_hash=$(echo "${deps_hash}\n$(cmake --version)" | sha256sum | cut -d " " -f 1)
+
         echo "hash=${deps_hash:0:9}" >> $GITHUB_OUTPUT
 
     - name: Restore Dependencies from Cache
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Looks good to me. CI failure is a different caching issue (VS/MSVC version mismatch between the cached files and the current run).

RytoEX:
As an aside, GitHub is in the middle of deploying a new Windows 2022 image, but neither the CMake version nor VS/MSVC version should have changed between it and the previous version.

PR #225, opened 2023-12-06
deps.ffmpeg: Fix fixup step error on windows

Chensiyy:
### Description

Fixed fixup step file path.

### Motivation and Context

The scripts failed when I execute them.

### How Has This Been Tested?

Successfully built with WSL on my computer.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/20-libpng.zsh
+++ b/deps.ffmpeg/20-libpng.zsh
@@ -145,7 +145,7 @@ fixup() {
         autoload -Uz create_importlibs
         create_importlibs ${target_config[output_dir]}/bin/libpng*.dll(:a)
         strip_tool=${target_config[cross_prefix]}-w64-mingw32-strip
-        strip_files=("${target_config[output_dir]}"/bin/bin/libpng*.dll(.))
+        strip_files=("${target_config[output_dir]}"/bin/libpng*.dll(.))
       } else {
         rm -rf -- ${target_config[output_dir]}/lib/libpng*.dll(N)
       }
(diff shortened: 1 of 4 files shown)


RytoEX:
The `bin/bin` here seems like an obvious mistake and probably should be corrected (despite that we no longer cross-compile for Windows). I'm not sure about the other changes. @PatTheMav ?

PatTheMav (approved):
Looks good to me - we'll hopefully not use these for Windows any longer, but as long as we keep cross-compilation as an emergency fallback, it's good to have this fixed.

RytoEX (changes requested):
Please add a space between each commit prefix and the rest of the commit message subject.
```
deps.ffmpeg: Fix
```

deps.ffmpeg/30-svt-av1.zsh
```diff
@@ -111,12 +111,11 @@ fixup() {
       ;;
     windows-x*)
       if (( shared_libs )) {
-        mv (#i)"${target_config[output_dir]}"/lib/libsvtav1*.dll ${target_config[output_dir]}/bin/
```
PatTheMav, line 114: On second look this change needs to be reverted - the `rm` command further below relies on the DLL to have an all-lowercase name, I also faintly remember other issues raised from the file names being mixed case.

PatTheMav: Also the DLL still needs to be moved to the `bin` directory.

Chensiyy: I have reverted this commit

Chensiyy:
> Please add a space between each commit prefix and the rest of the commit message subject.
> 
> ```
> deps.ffmpeg: Fix
> ```

I've fixed it.

RytoEX (approved):
Looks good to me.

PR #230, opened 2024-01-19
deps.ffmpeg: Fix MbedTLS compile error with Clang 15

PatTheMav:
### Description
Fixes compilation errors of MbedTLS when compiling with Clang 15.

### Motivation and Context
Clang 15 is the default compiler version on current macOS versions. The only reason this bug doesn't occur on CI is because the macOS 13 runners (and associated default Xcode version) are outdated.

### How Has This Been Tested?
Tested with Clang 15 on macOS 14.2.1.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/60-mbedtls.zsh
+++ b/deps.ffmpeg/60-mbedtls.zsh
@@ -73,7 +73,7 @@ config() {
     -DGEN_FILES=OFF
   )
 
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Seems fine for our purposes. It does have the limitation that if there is no CMAKE_C_FLAGS set in args, it will do nothing, but we currently always set those.

Will merge when I do a cleanup pass of obs-deps.

PR #232, opened 2024-01-23
utils.pwsh: Add support for curl resume

RytoEX:
### Description

Note that there is a bug in curl that can prevent curl from successfully interpreting a fully downloaded file as completed. The bug was fixed on October 30, 2023 and is not yet available on curl included with Windows.

Fixed in curl 8.5.0.
https://github.com/curl/curl/commit/225db9196a71a19132a4c334f256b59854afc70c

### Motivation and Context

I noticed that we have the Resume parameter, but that it wasn't implemented. I wanted to implement it to have a way to skip re-downloading a file that was already downloaded. SkipDeps skips build dependencies. SkipUnpack skips git checkouts and archive extraction but not downloads.

Alternatively, we could add something like SkipDownload and implement that, and remove the Resume parameter.

It may be important to note that due to us invoking MSYS2, the copy of curl from there can take precedence in PATH. From my recent tests, MSYS2 curl (8.1.1 on my machine) is older than the one shipped with Windows, so this may not work unless we attempt to fix the curl executable's path to the Windows copy (which should always be at `C:\Windows\System32\curl.exe`) or locate it and save its path before invoking MSYS2 (with `where.exe curl.exe`).

### How Has This Been Tested?

Tested locally with curl 8.4.0 (what is currently shipped on Windows) to ensure that the code did not currently run.

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Invoke-SafeWebRequest.ps1
+++ b/utils.pwsh/Invoke-SafeWebRequest.ps1
@@ -64,6 +64,11 @@ function Invoke-SafeWebRequest {
                 '--output', "$OutFile"
             )
 
+            $CurlVersionResult = $($(Invoke-External curl --version) -join ' ') -match 'curl (?<CurlVersion>\d+\.\d+\.\d+)'
+            if ( ( $Matches.CurlVersion -ge '8.5.0' ) -and $Resume ) {
+                $CurlOptions += @('-C', '-')
+            }
+
             Invoke-External curl @CurlOptions @HeaderStrings $Uri
 
             $NewHash = Get-FileHash -Path $OutFile -Algorithm $Algorithm
--- a/utils.pwsh/Invoke-SafeWebRequest.ps1
+++ b/utils.pwsh/Invoke-SafeWebRequest.ps1
@@ -57,7 +57,14 @@ function Invoke-SafeWebRequest {
                 }
             }
 
-            Invoke-External curl --fail --location $(if ( $Env:CI -eq $null ) { '--progress-bar' }) --output $OutFile @HeaderStrings $Uri
+            $CurlOptions = @(
+                '--fail'
+                '--location'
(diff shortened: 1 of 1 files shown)


RytoEX:
I checked today and currently Windows ships curl 8.7.1. The MSYS2 package via winget has also been updated and ships curl 8.7.1. As such, this would now work if we were interested in merging it.

utils.pwsh/Invoke-SafeWebRequest.ps1
```diff
@@ -57,7 +57,13 @@ function Invoke-SafeWebRequest {
                 }
             }
 
-            Invoke-External curl --fail --location $(if ( $Env:CI -eq $null ) { '--progress-bar' }) --output $OutFile @HeaderStrings $Uri
+            $ResumeOption = @()
```
PatTheMav, line 60: With now two conditional arguments to `curl`, I wonder if it weren't more readable if we compose one `CurlOptions` object and add all elements at once (including the progress bar and continue option) and then just call `Invoke-External curl @CurlOptions`.

RytoEX: That would probably be better. I can do that here or in a separate PR.

PatTheMav: I'd prefer making the "right" change immediately.

RytoEX: I've pushed a change that implements the suggestion in one commit and then adds support for curl resume in another commit.

RytoEX: Let me know if you'd like the code adjusted further.

RytoEX:
For posterity, I checked this locally with the qt package, and it seemed to work as expected.

PR #233, opened 2024-01-26
Replace AJA ntv2 repo with new libajantv2 repo

paulh-aja:
### Description

Swapped out usage of the old `ntv2` repo with the new `libajantv2` repo, using the latest 17.0.1 release tag: https://github.com/aja-video/libajantv2/commit/b6acce6b135c3d9ae7a2bce966180b159ced619f

The original `ntv2` repo used by the obs-studio `aja` capture and output plugins has been deprecated by AJA. This PR replaces the ntv2 repo deps with the new libajantv2 repo.

**New** - https://github.com/aja-video/libajantv2
Old - https://github.com/aja-video/ntv2

### Motivation and Context

The original open-source [**ntv2** repository](https://github.com/aja-video/ntv2) has been deprecated as it placed a cumbersome maintenance burden on the AJA software engineering team. That repository contained code that had to be filtered from an internal AJA code base. The [new libajantv2 repository](https://github.com/aja-video/libajantv2) separates that code from the AJA internal code base so that it can be used as a submodule in any project.

Updating to the new libajantv2 library is also required in order to support the new [AJA Kona X](https://www.aja.com/products/kona-x) board, released in early 2024.

### How Has This Been Tested?

Tested by building the libajantv2 obs-deps for Windows and macOS, via the provided scripts (Build-Dependencies.ps1 and build-deps.zsh, respectively). The built deps were then copied into the `.deps` dir of an obs-studio checkout, which was used to build the AJA OBS plugins on both Windows and macOS. Building obs-studio with the updated libajantv2 deps required a few changes which will go into a separate PR in the obs-studio repo.

### Types of changes

- Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:


deps.macos/20-libajantv2.zsh
```diff
@@ -34,10 +34,15 @@ config() {
...
-    -DAJA_BUILD_APPS=OFF
+    -DAJANTV2_DISABLE_DEMOS=ON
+    -DAJANTV2_DISABLE_DRIVER=ON
+    -DAJANTV2_DISABLE_TESTS=ON
+    -DAJANTV2_DISABLE_TOOLS=ON
+    -DAJANTV2_DISABLE_PLUGINS=ON
     -DAJA_INSTALL_SOURCES=OFF
     -DAJA_INSTALL_HEADERS=ON
+    -DAJA_INSTALL_MISC=OFF
+    -DAJA_INSTALL_CMAKE=OFF
```
PatTheMav, line 45: Add the `:BOOL` type hint here as well - we haven't added them to existing build scripts consistently, but when modifying them we might as well fix that while we're at it.

paulh-aja: I did not see those type hints used in the other .zsh build scripts so I wasn't sure if it was desired. I only saw them used in the PowerShell scripts.

paulh-aja: Added the type hints to the macOS libajantv2 script.

PatTheMav: > I did not see those type hints used in the other .zsh build scripts so I wasn't sure if it was desired. I only saw them used in the PowerShell scripts.

Yep that's fine, as I said that's just a nice to have so we don't have to modify the file yet another time just for that (because at some point we'll need to add them to all .zsh files as well).

deps.macos/20-libajantv2.zsh
```diff
@@ -1,10 +1,10 @@
...
 
 ## Dependency Information
-local name='ntv2'
-local version='16.2'
-local url='https://github.com/aja-video/ntv2.git'
-local hash='0acbac70a0b5e6509cca78cfbf69974c73c10db9'
+local name='libajantv2'
+local version='17.0'
+local url='https://github.com/aja-video/libajantv2.git'
+local hash='08a5a3e7e5fce4abf7deb11567dfd795da4a4ba0'
```
RytoEX, line 7: Update to 17.0.1.

deps.windows/60-libajantv2.ps1
```diff
@@ -1,8 +1,8 @@
 param(
-    [string] $Name = 'ntv2',
-    [string] $Version = '16.2',
-    [string] $Uri = 'https://github.com/aja-video/ntv2.git',
-    [string] $Hash = '0acbac70a0b5e6509cca78cfbf69974c73c10db9',
+    [string] $Name = 'libajantv2',
+    [string] $Version = '17.0',
+    [string] $Uri = 'https://github.com/aja-video/libajantv2.git',
+    [string] $Hash = '08a5a3e7e5fce4abf7deb11567dfd795da4a4ba0',
```
RytoEX, line 5: Update to 17.0.1.

RytoEX (changes requested):
Please update to the latest tag to (hopefully) resolve some issues. Will try to re-review/re-test against obs-studio before and after its companion PR.

paulh-aja:
> Please update to the latest tag to (hopefully) resolve some issues.

What issues are you referring to?

RytoEX:
> > Please update to the latest tag to (hopefully) resolve some issues.
> 
> What issues are you referring to?

IIRC, when I last tried to build obs-studio with the 17.0.0 tag, I ran into issues. I do not know if I have them documented.

RytoEX (changes requested):
Prefix the deps.windows commit with "deps.windows: " and the deps.macos commit with "deps.macos: ".

Please also treat the commit message subject and commit message body as separate strings/sentences.

RytoEX:
As far as I can tell, obs-studio at commit https://github.com/obsproject/obs-studio/commit/10bfa99365c1ed1df27f3a0af60c218a46d9071b does not build against these deps. This means that an update requiring this change would need to be done at the exact same time as code changes in obs-studio, which complicates this merge as, conversely, if we merge this, we cannot update the deps package _without_ also updating code in obs-studio.

paulh-aja:
> As far as I can tell, obs-studio at commit [obsproject/obs-studio@10bfa99](https://github.com/obsproject/obs-studio/commit/10bfa99365c1ed1df27f3a0af60c218a46d9071b) does not build against these deps. This means that an update requiring this change would need to be done at the exact same time as code changes in obs-studio, which complicates this merge as, conversely, if we merge this, we cannot update the deps package _without_ also updating code in obs-studio.

I updated the code in my [obs-studio PR](https://github.com/obsproject/obs-studio/pull/10146) to accommodate building with either the old ntv2 library or the new libajantv2 library. If the desire is to be able to merge this obs-deps PR without breaking the current obs-studio build, then we'd need to update this PR to keep building the old ntv2 library in addition to the new libajantv2 library (for a time). The only problem there is that the .lib/.a outputs that are produced by each build conflict with one another, as they have different names. The include paths don't conflict though, as they are stored in slightly different subdirectories.

RytoEX:
> > As far as I can tell, obs-studio at commit [obsproject/obs-studio@10bfa99](https://github.com/obsproject/obs-studio/commit/10bfa99365c1ed1df27f3a0af60c218a46d9071b) does not build against these deps. This means that an update requiring this change would need to be done at the exact same time as code changes in obs-studio, which complicates this merge as, conversely, if we merge this, we cannot update the deps package _without_ also updating code in obs-studio.
> 
> I updated the code in my [obs-studio PR](https://github.com/obsproject/obs-studio/pull/10146) to accommodate building with either the old `ntv2` library or the new `libajantv2` library. We would have to do something analogous here in obs-deps, if the desire is to be able to merge this obs-deps PR without breaking the current obs-studio build, and without needing to merge an obs-studio commit at the same time.
> 
> In that case we'd need to update this PR to keep building the old ntv2 library in addition to the new libajantv2 library (for a time). The only problem there is that the .lib/.a outputs that are produced by each build conflict with one another, as they have different names. The include paths don't conflict though, as they are stored in slightly different subdirectories.

Got it. If the obs-studio PR enables building with either library, then that's fine, and _that_ should be reviewed and merged first, not this.

Though, that PR is currently failing to build on Flatpak.

paulh-aja:
> Though, that PR is currently failing to build on Flatpak.

I'm not familiar with the Flatpak build. I was under the impression that we didn't build the AJA plugin on Linux. Looking at the logs for the failed Flatpak build in that PR, it appears that it is trying to find LibAJANTV2 but failing because it can't find it. Shouldn't the finder just not be executed on that build?

RytoEX (changes requested):
You should probably replace the license folder and file (remove them):
```
licenses/ntv2/LICENSE
```
with the current license in the new libajantv2 repo at:
```
licenses/libajantv2/LICENSE
```

Make sure that it gets copied appropriately with the new scripts.

Otherwise, this seems fine.

RytoEX (changes requested):
> Rename ntv2 license directory to libajantv2.
> 
> The original ntv2 library has been deprecated and
> replaced by libajantv2. The license is still MIT.

Never use final punctuation in the commit message subject (the first line of the commit message).

Wrap commit message bodies (the commit message lines that aren't the subject) at 72 characters, not 50.

I suggest being more explicit that this is not just a directory rename. Here is my suggested commit message:

> Replace ntv2 license with libajantv2 license
> 
> The original ntv2 library has been deprecated and replaced by
> libajantv2. The license is still MIT.

RytoEX (approved):
Looks good to me.

PatTheMav (approved):
Seems fine to me as well.

PR #241, opened 2024-03-14
CI: Force use of Xcode 15.2 to fix runtime crashes on older macOS

PatTheMav:
### Description
Force use of Xcode 15.2 in macOS-based runners.

### Motivation and Context
Xcode 15.0.1 had a bug which lead to binaries crashing on macOS 12 or older when compiled with weak definitions (FB13097713).

Mainly exhibited by use of Qt in OBS Studio: https://bugreports.qt.io/browse/QTBUG-119848

### How Has This Been Tested?
Needs to be tested by building OBS with updated dependencies and generated version of OBS run on macOS 12 or macOS 11.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -78,6 +78,10 @@ jobs:
       - name: Setup Environment
         id: setup
         run: |
+          print '::group::Enable Xcode 15.2'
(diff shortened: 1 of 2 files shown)


RytoEX (changes requested):
We should also force Xcode 15.2 in the scheduled workflow.

RytoEX (approved):

PR #243, opened 2024-04-08
deps.qt: Add Qt Designer

derrod:
### Description

Build Qt Designer with our Qt deps 

### Motivation and Context

We should have always have a readily available version of designer people can use without having to suffer through Qt's maintenance tool and download like 3 GB of Qt builds and/or Qt Creator which may or may not be built against an older Qt version.

Additionally this avoids issues where people use "whatever they had insatlled" which may not match the version OBS is using (on macOS/Windows at least) and make changes to the XML format that don't match the shipping version.

Should also help unblock https://github.com/obsproject/obs-studio/pull/10419

### How Has This Been Tested?

We'll see if we compile ey?

### Types of changes

### Checklist:

### DIFF
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -216,9 +216,9 @@ qt_add_submodules() {
           -DFEATURE_designer:BOOL=ON
           -DFEATURE_linguist:BOOL=OFF
           -DFEATURE_pixeltool:BOOL=OFF
+          -DFEATURE_qtattributionsscanner:BOOL=OFF
           -DFEATURE_qtdiag:BOOL=OFF
           -DFEATURE_qtplugininfo:BOOL=OFF
-          -DFEATURE_qtattributionsscanner:BOOL=OFF
           -DQT_BUILD_TOOLS_BY_DEFAULT:BOOL=ON
         )
       fi
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -16,6 +16,7 @@ local -a qt_components=(
   'qtshadertools'
   'qtmultimedia'
   'qtsvg'
+  'qttools'
 )
 
 local dir='qt6'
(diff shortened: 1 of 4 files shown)


derrod:
For some reason designer components were built, but the binary `designer.exe` was not created/installed, I guess this will require another dive into Qt Cmake...

RytoEX:
I did some digging and it looks like `-DQT_BUILD_TOOLS_BY_DEFAULT:BOOL=OFF` is the culprit here. We can try setting it back to ON just in the qttools configuration.

I filed [QTBUG-125067](https://bugreports.qt.io/browse/QTBUG-125067) in case this behavior is not intentional.

derrod:
Yeah setting that to `ON` just for `qttools` works, latest CI artifact has Qt Designer binaries.

deps.qt/qt6.zsh
```diff
@@ -215,6 +216,18 @@ qt_add_submodules() {
...
       if [[ ${component} == qtimageformats ]] _args+=(-DINPUT_tiff:STRING=qt -DINPUT_webp:STRING=qt)
+      if [[ ${component} == qttools ]]; then
+        _args+=(
+          -DFEATURE_assistant:BOOL=OFF
+          -DFEATURE_designer:BOOL=ON
+          -DFEATURE_linguist:BOOL=OFF
+          -DFEATURE_pixeltool:BOOL=OFF
+          -DFEATURE_qtdiag:BOOL=OFF
+          -DFEATURE_qtplugininfo:BOOL=OFF
+          -DFEATURE_qtattributionsscanner:BOOL=OFF
```
RytoEX, line 227:
```suggestion
          -DFEATURE_qtattributionsscanner:BOOL=OFF
          -DFEATURE_qtdiag:BOOL=OFF
          -DFEATURE_qtplugininfo:BOOL=OFF
```

RytoEX (changes requested):
Nit on configure option order. Besides that, and the additional time required to build this, the PR contents are fine.

RytoEX (approved):
The contents of the PR are fine. The additional build time is unfortunate, but I don't know how else we're going to have a version of Designer that matches our Qt version.

PR #250, opened 2024-05-04
deps.qt: Update more qt build options

RytoEX:
### Description

deps.qt: Fix libjpeg and libpng feature flags

These were apparently incorrect.

https://github.com/qt/qtbase/commit/3bc39a6026462e7e80694333c845ffe6d6b783ce
https://bugreports.qt.io/browse/QTBUG-120036

----

deps.qt: Remove unused CMake variables

Remove unused CMake variables to prevent this warning:
Manually-specified variables were not used by the project

FEATURE_printsupport gates:
FEATURE_cups
FEATURE_printdialog
FEATURE_printer
FEATURE_printpreviewdialog
FEATURE_printpreviewwidget

FEATURE_testlib gates:
FEATURE_itemmodeltester

### Motivation and Context

I was doing some local builds and kept seeing this warning:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    FEATURE_cups
    FEATURE_itemmodeltester
    FEATURE_libjpeg
    FEATURE_libpng
    FEATURE_printdialog
    FEATURE_printer
    FEATURE_printpreviewdialog
    FEATURE_printpreviewwidget
    FEATURE_system_libjpeg
    FEATURE_system_libpng
```

I wanted to see if we could prevent that.

Originally, when we worked on the CMake calls for configuring Qt, I think we intentionally opted to explicitly set these feature flags in an effort to convey intent (we want these disabled always). If we want to continue doing that and just bear with the warnings, that's fine. However, I think given that we have git history containing that these were set, we can probably check the git history to see that they were once explicitly set. If we decide that we still want to explicitly set them, I can just drop the second commit.

### How Has This Been Tested?

I did a local test build of Qt 6.6.3 on Windows 11. I compared its configure and build logs and the artifacts to those on CI.

### Types of changes

### Checklist:


PatTheMav (approved):
Seems fine, a cursory check of the CI build log seems to suggest that the features are not build with the gating variables set to `NO`.

RytoEX:
> Seems fine, a cursory check of the CI build log seems to suggest that the features are not build with the gating variables set to `NO`.

I forget how we log the configure summary for Qt, but I sometimes wonder if it would be helpful for us to cache Qt's configure summary and CMakeCache files for inspection later. I only discovered these when building locally because the CMake warning weren't emitted in the CI log. Probably more trouble than it's worth, but it's an idle thought.

PR #261, opened 2024-09-04
deps.ffmpeg: Remove MbedTLS CMake files from macOS FFmpeg package

PatTheMav:
### Description
Removes MbedTLS CMake package files for macOS FFmpeg builds.

### Motivation and Context
Once the project switches MbedTLS 3.6+, there needs to be a way to distinguish between older and newer versions on the obs-studio side.

The path of least resistance is to use CMake package-based discovery for versions 3.6+ and find module-based discovery for older versions, which requires the older (current as of time of the commit) versions to not provide a CMake package as part of obs-deps.

> [!NOTE]
> Once MbedTLS is updated to version 3.6+, this removal will need to be removed again.

### How Has This Been Tested?
Tested removal on current `obs-deps` with an updated `obs-studio` to opportunistically use package discovery mode and observe automatic fallback to find module to discover MbedTLS as provided by current dependencies.

### Types of changes

### Checklist:

### DIFF
--- a/build-deps.zsh
+++ b/build-deps.zsh
@@ -87,6 +87,7 @@ package() {
 
     rm -rf -- lib/^(*.dylib|*.a|*.so*|*.lib|*.framework|*.dSYM|cmake)(N)
     rm -rf -- lib/(libpcre*|libpng*|libfreetype.a)(N)
+    rm -rf -- lib/cmake/MbedTLS(N)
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Seems fine, though we'll have to remember to remove this when eventually updating mbedTLS.

PR #269, opened 2024-11-21
CI: Force-remove pkg-config on macOS runners and install pkgconf

PatTheMav:
### Description
Fix CI failures due to pre-installed `pkg-config` Homebrew formula blocking installation of `pkgconf`, which is aliased to `pkg-config` since an update to Homebrew which was released _after_ the new macOS runner images were created.

### Motivation and Context
Homebrew has effectively deprecated the pkg-config formula and instead aliased pkgconf to that name.

Installing pkg-config via the Brewfile thus leads to Homebrew attempting to install pkgconf instead, which will fail because the (older) pkg-config formula is installed by default on macOS runners.

### How Has This Been Tested?
Needs to be tested on CI.

### Types of changes

### Checklist:


.github/workflows/scheduled.yaml
```diff
@@ -99,19 +99,18 @@ jobs:
...
-          local -a to_remove=()
+          local -a unwanted_formulas=(llvm gcc pkg-config@0.29.2)
 
-          for formula (llvm gcc postgresql openjdk sox libsndfile flac libvorbis opusfile \
-            libogg composer php gd freetype fontconfig webp libpng lame libtiff opus kotlin \
-            sbt libxft libxcb) {
-            if [[ -d /usr/local/opt/${formula} ]] to_remove+=(${formula})
+          local -a remove_formulas=()
+          for formula (${unwanted_formulas}) {
+            if [[ -d ${HOMEBREW_PREFIX}/Cellar/${formula} ]] remove_formulas+=(${formula})
```
RytoEX, line 110: What's the reason for changing this list in this place but not the other?

.github/workflows/main.yaml
```diff
@@ -190,7 +191,7 @@ jobs:
 
   ffmpeg-package-universal:
     name: Build FFmpeg (Universal)
-    runs-on: macos-14
+    runs-on: macos-15
```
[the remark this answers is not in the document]
PatTheMav: According to a comment by a GitHub employee we should switch to it, as macOS 15 is about to become the default macOS image anyway.

.github/workflows/scheduled.yaml
PatTheMav: Copy+paste error I'd guess.

RytoEX:
Seems fine. Is the issue with png-config expected to arise again, or is this mostly preventative at this point?

RytoEX (approved):
Seems fine.

PR #270, opened 2024-12-15
CI: Include dSYMs in release for macOS dependencies

jcm93:
### Description

Adds a wildcard to the end of the filenames to be uploaded in the final release for macOS artifacts.

### Motivation and Context

In [releases](https://github.com/obsproject/obs-deps/releases/tag/2024-09-12), dSYMs for macOS dependencies are not being uploaded properly. Checksums are being generated, showing that the files are present. The Windows final deps upload paths include a wildcard at the end (to capture the `-PDBs` suffix), but this wildcard is not present for the macOS artifacts.

### How Has This Been Tested?

Tested on [local fork](https://github.com/jcm93/obs-deps/releases/tag/2024-12-23).

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -641,7 +641,7 @@ jobs:
             _temp=$(mktemp -d)
             pushd "${_temp}" > /dev/null
 
-            for artifact in ${GITHUB_WORKSPACE}/**/macos-@(deps|ffmpeg)-!(qt6*)-${arch}.*; do
+            for artifact in ${GITHUB_WORKSPACE}/**/macos-@(deps|ffmpeg)-!(qt6*)-${arch}?(-dSYMs).*; do
               case ${artifact} in
                 *.zip) unzip -o ${artifact} > /dev/null ;;
                 *.tar.xz) XZ_OPT=-T0 tar -xvJf ${artifact} ;;
@@ -649,8 +649,9 @@ jobs:
               esac
             done
 
-            XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz -- *
-            mv macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz ${GITHUB_WORKSPACE}
+            XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz -- !(*.dSYM)
+            XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}-dSYMs.tar.xz -- *.dSYM
+            mv macos-deps-${{ steps.metadata.outputs.version }}-${arch}?(-dSYMs).tar.xz ${GITHUB_WORKSPACE}
 
             popd > /dev/null
           done
@@ -678,6 +679,6 @@ jobs:
           files: |
             ${{ github.workspace }}/windows-*-x64*.zip
             ${{ github.workspace }}/windows-*-x86*.zip
-            ${{ github.workspace }}/macos-*-arm64.tar.xz
-            ${{ github.workspace }}/macos-*-x86_64.tar.xz
-            ${{ github.workspace }}/macos-*-universal.tar.xz
+            ${{ github.workspace }}/macos-*-arm64?(-dSYMs).tar.xz
+            ${{ github.workspace }}/macos-*-x86_64?(-dSYMs).tar.xz
+            ${{ github.workspace }}/macos-*-universal?(-dSYMs).tar.xz


.github/workflows/main.yaml
```diff
@@ -678,6 +678,6 @@ jobs:
           files: |
             ${{ github.workspace }}/windows-*-x64*.zip
             ${{ github.workspace }}/windows-*-x86*.zip
-            ${{ github.workspace }}/macos-*-arm64.tar.xz
-            ${{ github.workspace }}/macos-*-x86_64.tar.xz
-            ${{ github.workspace }}/macos-*-universal.tar.xz
+            ${{ github.workspace }}/macos-*-arm64*.tar.xz
+            ${{ github.workspace }}/macos-*-x86_64*.tar.xz
+            ${{ github.workspace }}/macos-*-universal*.tar.xz
```
PatTheMav, line 683: I'd prefer a more exclusive glob expression here, e.g. `macos-*-arm64?(-dSYMs).tar.xz`.

The GitHub Actions runtime probably uses `node-glob` which explicitly supports Bash' `extglob` patterns.

PatTheMav: The `softprops/action-gh-release` uses the `glob` npm package, which does.

This is not a workflow glob, but an input to a Node-based action.

jcm93: Changed; made sure the glob expression worked out-of-band and it seems to.

PatTheMav (commented):
Looks fine to me.

@jcm93 Did you run a test release on your fork to check if the glob works as intended?

jcm93:
Yes; I linked it in the earlier conversation: https://github.com/jcm93/obs-deps/releases/tag/2024-12-23

PatTheMav:
> Yes; I linked it in the earlier conversation: https://github.com/jcm93/obs-deps/releases/tag/2024-12-23

But there's only dSYMs for Qt6 it seems?

jcm93:
It looks like those are not among the downloaded artifacts for the release flow in the first place, looking at the checksums there and in the [latest official release](https://github.com/obsproject/obs-deps/releases/tag/2024-09-12). dSYMs are not present there nor are PDBs for Windows. I had actually assumed that that was intentional and there was some other reason debug symbols weren't available. If we want I can expand the scope of this to figure out why dSYMs and PDBs aren't included for the main prebuilt batch.

RytoEX:
> It looks like the missing dSYMs (for the generic `macos-deps-2024-12-23-arm64.tar.xz` etc.) are not among the downloaded artifacts for the release flow in the first place, looking at the checksums there and in the [latest official release](https://github.com/obsproject/obs-deps/releases/tag/2024-09-12). dSYMs are not present there nor are PDBs for the Windows counterparts (`windows-deps-2024-12-23-x64.zip` etc). I had actually assumed that that was expected and there was some other reason those debug symbols weren't available. If we want I can expand the scope of this to figure out why dSYMs and PDBs aren't included for the main prebuilt batch.

We have historically never enabled PDBs for the non-Qt Windows dependencies. That is a separate item to address (which required no longer cross-compiling many of the deps) and need not be part of this PR.

PatTheMav:
IMO this PR is incomplete if only the Qt6 dSYMs are made available in a release (as all the other dSYMs are generated as artefacts). Because FFmpeg and main dependencies are smushed together into a new archive, you'd probably have to extend the workflow step to do the same for the dSYM archives as well.

jcm93:
I'll take a look at getting dSYMs for everything else.

RytoEX:
> IMO this PR is incomplete if only the Qt6 dSYMs are made available in a release (as all the other dSYMs are generated as artefacts). Because FFmpeg and main dependencies are smushed together into a new archive, you'd probably have to extend the workflow step to do the same for the dSYM archives as well.

My main point about the Windows devs is that we do not even _generate_ PDBs for the FFmpeg dependencies on Windows yet. Addressing that here would be out of scope. If PDBs for everything else are already generated, but not included, then yes, it is fine to address that here.

PatTheMav:
> > IMO this PR is incomplete if only the Qt6 dSYMs are made available in a release (as all the other dSYMs are generated as artefacts). Because FFmpeg and main dependencies are smushed together into a new archive, you'd probably have to extend the workflow step to do the same for the dSYM archives as well.
> 
> My main point about the Windows deps is that we do not even _generate_ PDBs for the FFmpeg dependencies on Windows yet. Addressing that here would be out of scope. If PDBs for everything else are already generated, but not included, then yes, it is fine to address that here.

Yep the PDBs are a different matter entirely. For macOS all dependencies flavours generate dSYMs already though.

jcm93:
Updated; added a new step that packages the debug symbols, since it uses a separate glob expression. Test release is [here](https://github.com/jcm93/obs-deps/releases/tag/2025-01-20).

PatTheMav (commented):
There is no need for an additional step, this should do just fine:

```Bash
for arch in arm64 x86_64 universal; do
  _temp=$(mktemp -d)
  pushd "${_temp}" > /dev/null
  for artifact in ${GITHUB_WORKSPACE}/**/macos-@(deps|ffmpeg)-!(qt6*)-${arch}?(-dSYMs).*; do
    case ${artifact} in
      *.zip) unzip -o ${artifact} > /dev/null ;;
      *.tar.xz) XZ_OPT=-T0 tar -xvJf ${artifact} ;;
      *.tar.gz) tar -xvzf ${artifact} ;;
    esac
  done
  XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz -- !(*.dSYM)
  XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}-dSYMs.tar.xz -- *.dSYM
  mv macos-deps-${{ steps.metadata.outputs.version }}-${arch}?(-dSYMs).tar.xz ${GITHUB_WORKSPACE}
  popd > /dev/null
done
```

jcm93:
Alright yeah that seems to have worked. Pushed; test release [here](https://github.com/jcm93/obs-deps/releases/tag/2025-01-20).

.github/workflows/main.yaml
```diff
@@ -640,18 +640,16 @@ jobs:
           for arch in arm64 x86_64 universal; do
             _temp=$(mktemp -d)
             pushd "${_temp}" > /dev/null
-
```
RytoEX, line 643: Can probably leave this blank line intact.

.github/workflows/main.yaml
```diff
@@ -640,18 +640,16 @@ jobs:
...
               case ${artifact} in
                 *.zip) unzip -o ${artifact} > /dev/null ;;
                 *.tar.xz) XZ_OPT=-T0 tar -xvJf ${artifact} ;;
                 *.tar.gz) tar -xvzf ${artifact} ;;
               esac
             done
-
-            XZ_OPT=-T0 tar -cvJf macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz -- *
-            mv macos-deps-${{ steps.metadata.outputs.version }}-${arch}.tar.xz ${GITHUB_WORKSPACE}
-
```
RytoEX, line 654: Same for these blank lines.

RytoEX (changes requested):

.github/workflows/main.yaml
jcm93: Addressed.

.github/workflows/main.yaml
jcm93: Addressed.

PatTheMav (approved):
Looks good to me - obviously depends on the artefact contents staying as-is, but the entire CI infrastructure depends on every file fulfilling some expectation like that. 😅

RytoEX (approved):

PR #273, opened 2025-01-25
CI: Remove superfluous step to publish Windows Qt artifacts

PatTheMav:
### Description
Removes the publish step from the Windows Qt6 build job.

### Motivation and Context
Publish step was not intended to be contained in the build job and is a left-over from when the job was copied over from the main workflow.

### How Has This Been Tested?
Will need testing on CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -375,13 +375,6 @@ jobs:
           target: ${{ matrix.target }}
           config: ${{ matrix.config }}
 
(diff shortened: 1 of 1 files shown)


RytoEX:
Is it not needed so that the Qt WoA job has host components?

PatTheMav:
> Is it not needed so that the Qt WoA job has host components?

D'oh, you're right. I just spotted those two artefacts (and no others) and was confused by that.

PR #275, opened 2025-01-29
deps.windows: Fix architecture mismatch when building on ARM64 host

PatTheMav:
### Description
Changes the DevShell arguments to explicitly include the host architecture whenever the `x86` architecture needs to be built on GitHub Actions runners.

### Motivation and Context
While working without issue on local machines, it seems to be that the build scripts cannot correctly invoke a DevShell and the detours `nmake` build system for the x86 architecture without explicitly specifying the host architecture.

The same workaround breaks building other architecture on actual ARM64 Windows machines, thus it is now limited to CI runs only.

### How Has This Been Tested?
Tested locally in a Windows ARM64 VM, will be tested by CI for other architectures.

### Types of changes

### Checklist:

### DIFF
--- a/deps.windows/40-detours.ps1
+++ b/deps.windows/40-detours.ps1
@@ -40,7 +40,7 @@ function Build {
         Target = $Target
     }
 
-    if ( $Target -eq 'x86' ) {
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Looks fine.

PR #278, opened 2025-03-03
deps.ffmpeg: Update SVT-AV1 to 3.0.0

RytoEX:
### Description

Update SVT-AV1 to 3.0.0.

SVT-AV1 3.0.0 no longer vendors cpuinfo (see [SVT-AV1 release notes](https://gitlab.com/AOMediaCodec/SVT-AV1/-/releases/v3.0.0) and [the commit in question](https://gitlab.com/AOMediaCodec/SVT-AV1/-/commit/54ac9322e79508e6ef3474d77f8e32e4ca49b923)). They now use CMake's FetchContent by default, unless we specify at configure time to use an external copy. This means that we cannot Patch cpuinfo before Configure.

If we determine that the patch is no longer required, then we can simply remove it.

If we determine that the patch is still required, then we have to either:
 A. upstream our patch, then get SVT-AV1 to update the commit they use
    for FetchContent
 B. figure out if we can use a SYSTEM/EXTERNAL cpuinfo
 C. disable cpuinfo (USE_CPUINFO=OFF)

### Motivation and Context

Want to explore keeping dependencies up to date while managing our patches.

### How Has This Been Tested?

Built locally on Windows 11, and I think I built OBS Studio against that though I do not have clear notes on it. I'd like further review to try to understand what choices we have for dealing with the new cpuinfo changes.

### Types of changes

### Checklist:


RytoEX:
It looks like FFmpeg 7.1 doesn't support SVT-AV1 3.0.0, so this will have to wait for a future update. However, my questions about our cpuinfo patch remain, and we should probably think about those in advance.

PatTheMav:
> It looks like FFmpeg 7.1 doesn't support SVT-AV1 3.0.0, so this will have to wait for a future update. However, my questions about our cpuinfo patch remain, and we should probably think about those in advance.

I just checked their `FetchContent` implementation and also the commit of `cpuinfo` they use and indeed it seems as if the latter added first-class support for Visual Studio as a generator and for `x86`, `x64`, and several `ARM64` variants, so the patch is not needed anymore.

PatTheMav (approved):
Changes themselves look good, the FFmpeg incompatibility has been mentioned by Tytan already.

RytoEX:
Will probably have to wait for a new stable tag of FFmpeg before we can merge this, but glad to have the review.

RytoEX:
> Now that FFmpeg 7.1.1 has been merged, will this PR be reviewed?

No, because 7.1.1 does not contain the fix commit. We will re-review this when there is a stable tag that does.

he6ot:
@RytoEX Is it possible to update svt-av1 to version 4.1.0?

PR #280, opened 2025-03-03
deps.ffmpeg: Enable PDBs for FFmpeg builds on Windows

RytoEX:
### Description

Enable PDBs for FFmpeg builds on Windows if the Configuration is set to Debug or RelWithDebInfo.

### Motivation and Context

Having PDBs for FFmpeg can be useful for debugging crashes. This was one of the original goals of moving the FFmpeg builds to be natively compiled.

### How Has This Been Tested?

Tested locally about a year ago. Should probably be retested.

### Types of changes

### Checklist:

### DIFF
--- a/deps.ffmpeg/99-ffmpeg.ps1
+++ b/deps.ffmpeg/99-ffmpeg.ps1
@@ -68,7 +68,8 @@ function Configure {
         '--toolchain=msvc'
         ('--extra-cflags=' + "'-D_WINDLL -MD -D_WIN32_WINNT=0x0A00" + $(if ( $Target -eq 'arm64' ) { ' -D__ARM_PCS_VFP' }) + "'")
         ('--extra-cxxflags=' + "'-MD -D_WIN32_WINNT=0x0A00'")
-        ('--extra-ldflags=' + "'-APPCONTAINER:NO -MACHINE:${Target}'")
+        ('--extra-ldflags=' + "'-APPCONTAINER:NO -MACHINE:${Target} -DEBUG" +
+            $(if ( $Configuration -match '(Release|RelWithDebInfo|MinSizeRel)' ) { ' -OPT:REF -OPT:ICF -LTCG -INCREMENTAL:NO' }) + "'")
         $(if ( $Target -eq 'arm64' ) { '--as=armasm64.exe','--cpu=armv8' })
         '--pkg-config=pkg-config'
         $(if ( $Target -ne 'x86' ) { '--target-os=win64' } else { '--target-os=win32' })
--- a/deps.ffmpeg/99-ffmpeg.ps1
+++ b/deps.ffmpeg/99-ffmpeg.ps1
@@ -166,4 +166,9 @@ function Install {
     $env:VERBOSE = $(if ( $VerbosePreference -eq 'Continue' ) { '1' })
     Invoke-DevShell @Params
     $Backup.GetEnumerator() | ForEach-Object { Set-Item -Path "env:\$($_.Key)" -Value $_.Value }
+
+    Push-Location "build_${Target}"
+    Copy-Item 'ffmpeg_g.exe','ffprobe_g.exe' "$($script:ConfigData.OutputPath)/bin"
+    Get-ChildItem -Path '.' -Filter '*.pdb' -Recurse | Copy-Item -Destination "$($script:ConfigData.OutputPath)/bin"
+    Pop-Location
 }
--- a/Build-Dependencies.ps1
+++ b/Build-Dependencies.ps1
@@ -122,7 +122,7 @@ function Package-Dependencies {
 
     switch ( $PackageName ) {
         ffmpeg {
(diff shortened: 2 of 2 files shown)


deps.ffmpeg/99-ffmpeg.ps1
```diff
@@ -67,7 +67,7 @@ function Configure {
         '--toolchain=msvc'
         ('--extra-cflags=' + "'-D_WINDLL -MD -D_WIN32_WINNT=0x0A00" + $(if ( $Target -eq 'arm64' ) { ' -D__ARM_PCS_VFP' }) + "'")
         ('--extra-cxxflags=' + "'-MD -D_WIN32_WINNT=0x0A00'")
-        ('--extra-ldflags=' + "'-APPCONTAINER:NO -MACHINE:${Target}'")
+        ('--extra-ldflags=' + "'-APPCONTAINER:NO -MACHINE:${Target}'" + $(if ( $Configuration -match '(Debug|RelWithDebInfo)' ) { ' -DEBUG' }))
```
PatTheMav, line 70: Wouldn't it make sense to always add this flag? Calling PDBs "debug"-symbols is a bit of a misleading name, as "debug" is their purpose but is not tied to the build configuration. It is perfectly fine (and indeed expected) to generate PDB symbol files in `Release` configuration (or _particularly_ for that configuration) to ship stripped release binaries to users while still being able to debug/symbolicate issues.

RytoEX: We could. My recollection from our conversation when I first wrote this a year ago was that blanket adding `-DEBUG` all the time was a quick hack and that we should only add it as needed. If our current consensus is that we should always use this flag, then we can do that.

RytoEX: The default configuration for `.github/actions/build-ffmpeg/action.yaml` is `Release` and it seems that we currently specify that in `.github/workflows/main.yaml` and `.github/workflows/scheduled.yaml`.

PatTheMav: That was my hunch - so if we actually want to have PDBs generated for `obs-deps` releases, then the `DEBUG` switch needs to be probably added regardless of selected build configuration.

This will work for CMake-based projects using the MSVC build system, but I'm not sure about custom build systems like FFmpeg's.

RytoEX: How do we want to proceed here?

RytoEX: I've pushed an implementation as [written above](https://github.com/obsproject/obs-deps/pull/280#discussion_r2096576063):
```pwsh
        ('--extra-ldflags=' + "'-APPCONTAINER:NO -MACHINE:${Target} -DEBUG" +
            $(if ( $Configuration -match '(Release|RelWithDebInfo|MinSizeRel)' ) { ' -OPT:REF -OPT:ICF -LTCG -INCREMENTAL:NO' }) + "'")
```

If we want to expand this to `Setup-BuildParameters`, we can make that change later.

RytoEX:
Just confirming that the recent push still creates PDBs for the DLLs:
image

Executables are suffixed for some reason:
image

The PDBs are not currently copied/installed to the install dir.

RytoEX:
Apparently, I had forgotten to actually push the changes. The PDBs are still not actually part of any final artifacts, as I have to figure out where those are installed/copied.

PatTheMav (approved):

PatTheMav:
Looking at my own code from the distance of a few years makes me want to refactor the entire build system for `obs-deps`, but that's a different story.. 😅

deps.ffmpeg/99-ffmpeg.ps1
```diff
@@ -165,4 +166,9 @@ function Install {
     $env:VERBOSE = $(if ( $VerbosePreference -eq 'Continue' ) { '1' })
     Invoke-DevShell @Params
     $Backup.GetEnumerator() | ForEach-Object { Set-Item -Path "env:\$($_.Key)" -Value $_.Value }
+
+    Push-Location "build_${Target}"
+    Copy-Item "ffmpeg_g.exe","ffprobe_g.exe" "$($script:ConfigData.OutputPath)/bin"
```
PatTheMav: It seems that I chose to use single quotes for plain string tokens and only used double quotes when interpolation is necessary.

I dunno if the code is consistent that way, but at least the lines in `Build-Dependencies.ps1` suggest that pattern is in use.

RytoEX: I don't think our code is consistent about it. The documentation for `Copy-Item` uses double quotes for nearly everything, and I was referencing that at the time. I can change it.

RytoEX:
Following up: PDBs are now manually copied to the install dir and are made part of the `ffmpeg-windows-arm64` and `
ffmpeg-windows-x64` artifacts instead of being in separate artifacts. May change this later to split them out, but they add fairly little to the existing archive sizes (4~6 MB).

PR #281, opened 2025-03-03
winget: Fix winget meson installation

RytoEX:
### Description

Fix winget meson installation. Without this change, meson will not be installed automatically via winget. Can change the first commit prefix to `utils.pwsh` if preferred, since it also modifies a file there.

A second commit changes the syntax for `.Wingetfile` so that it is uniform for all entries.

### Motivation and Context

Pointed out on #277. Further investigation showed that this does indeed fail if meson cannot be found with `Get-Command`.

### How Has This Been Tested?

Tested locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/.Wingetfile
+++ b/.Wingetfile
@@ -1,4 +1,4 @@
 package '7zip.7zip', path: '7-zip', bin: '7z'
 package 'cmake', path: 'Cmake\bin', bin: 'cmake'
-package 'MSYS2.MSYS2', path: 'C:\msys64\usr\bin', bin 'bash'
-package 'mesonbuild.meson', path: 'Meson', bin 'meson'
+package 'MSYS2.MSYS2', path: 'C:\msys64\usr\bin', bin: 'bash'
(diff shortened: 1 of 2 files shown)


RytoEX:
Hmm. Odd that the WoA builds failed with this change. Will have to investigate further.

.Wingetfile
PatTheMav, line 1: Syntactically it should still be `package <id>[, path: <path>[, bin: <binary>]]`, as the `package` keyword starts a package declaration, whereas `path` or `bin` are attributes.

RytoEX: Fixed.

RytoEX:
Separately, I'm not sure why meson is special-cased to use Python on CI. Maybe winget didn't have a new enough version at the time or something, but I do not know for sure.

PatTheMav:
> Separately, I'm not sure why meson is special-cased to use Python on CI. Maybe winget didn't have a new enough version at the time or something, but I do not know for sure.

There is no winget on CI.

PatTheMav (approved):

PR #283, opened 2025-03-20
deps.qt: Enable creation of dSYM files for all configurations

PatTheMav:
### Description
Enable creation of separate debug information for all CMake configurations.

### Motivation and Context
On macOS debug information is kept in the object files generated by the compiler and extracted into dSYM files using the reference information kept in the binaries.

Any binary built in debug information will not contain any debug information, instead the toolchain will find the information in the object files locally. Thus binaries created in debug configuration need to be shipped with their associated dSYM files if they are to be used on a different machine.

### How Has This Been Tested?
Tested with a Qt build with debug configuration, used on a separate computer and confirmed that debug information was correctly found when the associated dSYM files were loaded.

### Types of changes

### Checklist:

### DIFF
--- a/deps.qt/qt6.zsh
+++ b/deps.qt/qt6.zsh
@@ -103,7 +103,7 @@ config() {
     fi
   }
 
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Seems fine.

PR #284, opened 2025-04-14
Update dependencies to ensure CMake 4 compatibility

PatTheMav:
### Description
Updates all affected dependencies to ensure compatibility with CMake 4.0.0, which has removed for any CMake script with a policy version below 3.5.

This deprecation has been forewarned since version 3.27, alas many projects have been entirely inactive or have not merged available PRs to either add or update their CMake support.

* srt: https://github.com/Haivision/srt/pull/3135
* Libdatachannel: https://github.com/paullouisageneau/libdatachannel/issues/1366
* Swig: Master branch has updated CMakeLists.txt, would require proper update of Swig to a more recent build, which is a more extensive update due to the stable ABI requirement
* Websocketpp: https://github.com/zaphoyd/websocketpp/pull/1168
* RNnoise: Has at least two open PRs to add CMake support, both of which seem to have met little interest yet

### Motivation and Context
Fix breaking `obs-deps` runs on CI, whose runners will use CMake 4.0.0 (or later) either soon or do so already.

### How Has This Been Tested?
Preliminary tests done on macOS 15 and Windows 11.

### Types of changes

### Checklist:


RytoEX:
> * Swig: Master branch has updated CMakeLists.txt, would require proper update of Swig to a more recent build, which is a more extensive update due to the stable ABI requirement

There are two options for updating SWIG.
1. If we want to update SWIG, we can (probably) easily revert back to using the prebuilt SWIG. However, there is no WoA prebuilt version at this time.
2. If we want to try to build a newer version of SWIG, that has separate issues.
New SWIG requires Bison 3.5+ (we're using a much older version).
The source we use for pre-compiled Bison doesn't have 3.5+.
A newer source for pre-compiled Bison (https://github.com/lexxmark/winflexbison) doesn't drop-in cleanly.

If we wanted to update SWIG now as option 1, we could do it if we dropped WoA SWIG.

I'm not super fond of updating dependencies to use git commits where we'd previously used stable tags, but there's not much else I think we can do.

RytoEX (changes requested):
Nits:
* srt -> SRT
* RNnoise -> RNNoise
* Swig -> SWIG

Will get to a full review later.

deps.windows/50-pcre.ps1
```diff
@@ -25,7 +25,6 @@ function Configure {
 
     $Options = @(
         $CmakeOptions
-        "-DBUILD_SHARED_LIBS:BOOL=OFF"
```
RytoEX, line 28: What's the reason here?

RytoEX (changes requested):

RytoEX:
For posterity, updates:
* deps.ffmpeg: [Update libogg from 1.3.5 to fa80aae9d50096160f2b56ada35527d7aee3f746](https://github.com/xiph/ogg/compare/v1.3.5...fa80aae9d50096160f2b56ada35527d7aee3f746)
* deps.ffmpeg: [Update libvorbis from 1.3.7 to 2eac96b03ff67953354cb0a649c08aa3a23267ef on macOS](https://gitlab.xiph.org/xiph/vorbis/-/compare/v1.3.7...2eac96b03ff67953354cb0a649c08aa3a23267ef?from_project_id=46)
* deps.macos: [Update jansson from 2.14 to 96d160df90016066d04d493d1d69639474ba4f20](https://github.com/akheron/jansson/compare/v2.14...96d160df90016066d04d493d1d69639474ba4f20)
* deps.macos: [Update pcre2 from 10.40 to 10.45](https://github.com/PCRE2Project/pcre2/compare/pcre2-10.40...pcre2-10.45)
* deps.windows: [Update jansson from 2.14 to 96d160df90016066d04d493d1d69639474ba4f20](https://github.com/akheron/jansson/compare/v2.14...96d160df90016066d04d493d1d69639474ba4f20)
* deps.windows: [Update pcre2 from 10.40 to 10.45](https://github.com/PCRE2Project/pcre2/compare/pcre2-10.40...pcre2-10.45)

deps.windows/50-pcre.ps1
PatTheMav: Looks like that one sneaked in through a cherry-pick, removed it with a force-push.

RytoEX (approved):

PatTheMav:
> This updates libvorbis for macOS but not for Windows, and seems to be the only case where a dependency is updated only for macOS. Is that intentional?

Good catch, that was also missed during the conversion from my other branch - added the change back.

PR #285, opened 2025-04-24
Add qtserialport to obs-deps

glikely:
This is a request to add another Qt component to obs-deps

The obs-ptz plugin uses qtserialport to communicate with remote PTZ cameras. Unfortunatly, it does not seem possible to build QSerialPort outside of the Qt6 library build environment because it links into private Qt interfaces. The plugin used to maintain a private fork of obs-deps just to add qtserialport, but that approach proved to be fragile. Some users have had success putting a copy of qtserialport from the official Qt binaries into the plugins directory, but again it is easy to get out of sync.

Adding qtserialport to the list of Qt modules built in obs-deps solves the problem with low complexity. The library is an official Qt component, is small, and doesn't pull in any additional dependencies.

@PatTheMav you may remember, I made the same request a couple of years ago. At the time you weren't keen to add to components due to long github action build times. Instead I tried maintaining a fork of obs-deps, but that was quite fragile as described above. The plugin became very hard to up rev and is broken since the 29.0 release.

I did some build time comparisons. Here are two builds, the first is a mainline commit, and the second is after adding qtserialport.

https://github.com/obsproject/obs-deps/actions/runs/14585093923/usage
https://github.com/glikely/obs-deps/actions/runs/14631880990/usage
Wall clock time: 1h 8m 09s   vs.  1h 8m 28s.  delta 19s
Usage time: 4h 18m 46s.  vs.   4h 21m 33s. delta 2m 47s

### Checklist:

- [n/a] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).

### DIFF
--- a/deps.qt/checksums/qtserialport-everywhere-src-6.8.2.tar.xz.sha256
+++ b/deps.qt/checksums/qtserialport-everywhere-src-6.8.2.tar.xz.sha256
@@ -1,0 +1,1 @@
+ed17e02361e989f149f58d021ab8cd66f21db4cdfb8cde0a462017e6ac1e3be7  qtserialport-everywhere-src-6.8.2.tar.xz
--- a/deps.qt/checksums/qtserialport-everywhere-src-6.8.2.zip.sha256
+++ b/deps.qt/checksums/qtserialport-everywhere-src-6.8.2.zip.sha256
@@ -1,0 +1,14 @@
+<Objs Version="1.1.0.1" xmlns="http://schemas.microsoft.com/powershell/2004/04">
+  <Obj RefId="0">
+    <TN RefId="0">
+      <T>Microsoft.PowerShell.Commands.FileHashInfo</T>
+      <T>System.Object</T>
+    </TN>
+    <ToString>Microsoft.PowerShell.Commands.FileHashInfo</ToString>
+    <Props>
+      <S N="Algorithm">SHA256</S>
+      <S N="Hash">BA061FC930D8DCEB9691C3273EA0FDDD6E1013409E7DD627F319B594D4EDDC2C</S>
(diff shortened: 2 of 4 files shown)


RytoEX:
I'm inclined to close this. As explained in https://github.com/obsproject/obs-deps/pull/137, obs-deps is for building dependencies that OBS Studio itself needs. It is not meant to build dependencies that all possible plugins could need. My reservation is not about build times on our CI but about setting precedent for adding dependencies for third-party plugins and adding maintenance burden.

> Unfortunatly, it does not seem possible to build QSerialPort outside of the Qt6 library build environment because it links into private Qt interfaces.

If you should be able to build QSerialPort outside of a Qt6 build environment, then I would suggest filing a QTBUG.

In the meantime, you should just be able to use Qt6SerialPorts from the Qt Installer, or locally build Qt with Qt Serial Port enabled.

Fenrirthviti:
Agree on the above, closing.

glikely:
Hi @RytoEX. I forgot that I was talking to you about this rather than PatTheMav. Sorry.

I was going to respond by arguing that in the Qt case it is reasonable to ask for additional submodules because building them outside of the full Qt build environment is difficult and that plugins shipping out of sync versions cause hard to track down bugs. Then I was going to share the battle scars and frustrations of trying to build and ship submodules via forking obs-deps.... but I was wrong.

Yes, the submodules build just fine without the Qt core source. I've got it working now on both MacOS and Windows. Either Qt6 has gotten much easier to work with, or I simply didn't know what I was doing the last time I tried. The fact that the updated obs-plugintemplate is much easier to work with certainly helped too.

Anyway, yes I now have a solution. Thanks.

PR #245, opened 2024-04-26
Update dependencies

RytoEX:
### Description

Update dependendies as available:
* deps.ffmpeg: Update libpng to 1.6.43
* deps.ffmpeg: Update opus to 1.5.2
* deps.ffmpeg: Update libvpx to v1.14.0
* deps.ffmpeg: Update SVT-AV1 to 2.0.0
* deps.ffmpeg: Update aom to 3.9.0
* deps.ffmpeg: Update libdatachannel to v0.21.0
* deps.ffmpeg: Update AMF to 1.4.33
* deps.ffmpeg: Update FFmpeg to 7.0
* deps.macos: Update LuaJIT to 2.1 5790d25397
* deps.macos: Update libpng to 1.6.43
* deps.macos: Update Asio to 1.30.2
* deps.macos: Update Sparkle to 2.6.1
* deps.windows: Update curl to 8.7.1
* deps.windows: Update LuaJIT to 2.1 5790d25397
* deps.windows: Update Asio to 1.30.2
* deps.windows: Update Zstandard to 1.5.6
* deps.windows: Update VPL to v2.10.2
* deps.qt: Update Qt6 to 6.6.3 for Windows
* deps.qt: Update Qt6 to 6.6.3 for macOS

### Motivation and Context

Want to keep our deps up to date.

Supersedes #246
Closes #246

Note that neither SRT/libsrt nor mbedTLS are updated here. Building SRT 1.5.x against mbedTLS 3.5.0+ fails because of changes in mbedTLS that needed addressed in libsrt, but there is no new libsrt tag with the needed changes.
https://github.com/Haivision/srt/pull/2842

I initially updated RNNoise, but dropped it because the Windows build actually uses an old PR that added CMake support. We'll either have to rewrite CMake support for the latest stable tag, or we'll have to change our build script to use their preferred build system (autotools).

I did not update librist because versions 0.2.8-0.2.10 have a bug, and there is no new tag. See:
* https://github.com/obsproject/obs-deps/pull/228#issuecomment-1897611888

Will post test notes in a follow-up comment.

### How Has This Been Tested?

Built deps locally. Built OBS against those deps on Windows 11.

### Types of changes

- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
[...]

PR #228, opened 2024-01-18
Update dependencies

RytoEX:
### Description

Update many of our dependencies.

* deps.ffmpeg: Update zlib to 1.3
* deps.ffmpeg: Update SVT-AV1 to 1.8.0
* deps.ffmpeg: Update aom to 3.8.0
* deps.ffmpeg: Update libsrt to 1.5.3
* deps.ffmpeg: Update nv-codec-headers to 12.1.14.0
* deps.ffmpeg: Update AMF to 1.4.32
* deps.ffmpeg: Update FFmpeg to 6.1
* deps.macos: Update LuaJIT to 2.1 c525bcb902
* deps.macos: Update FreeType to 2.13.2
* deps.macos: Update Asio to 1.29.0
* deps.macos: Update nlohmann JSON to 3.11.3
* deps.macos: Update Sparkle to 2.5.2
* deps.macos: Update Syphon Framework to 5.0 39e31383ff
* deps.windows: Update FreeType to 2.13.2
* deps.windows: Update curl to 8.5.0
* deps.windows: Update LuaJIT to 2.1 c525bcb902
* deps.windows: Update Asio to 1.29.0
* deps.windows: Update nlohmann JSON to 3.11.3
* deps.windows: Update VPL to v2.10.1
* deps.qt: Update Qt6 to 6.6.1 for Windows
* deps.qt: Update Qt6 to 6.6.1 for macOS

### Motivation and Context

Want to keep our deps up to date. Up to date deps come with bug fixes, new features, and allow us to maintain less patches and backport less commits/patches.

After the recent enablement of native Windows FFmpeg builds, this updates dependency versions to help provide a distinct comparison to that tag/release.

### How Has This Been Tested?

Tested locally. Will post results in a separate comment.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #289, opened 2025-06-03
Update librist to commit 186bedbb6a2ec77f994689ccdde4a9bd88548d22

lescurej:
### Description
Updated the dependency scripts `deps.ffmpeg/70-librist.zsh` and `deps.ffmpeg/70-librist.ps1` to reflect the latest commit hash of `librist`.

This ensures OBS builds with the most recent version of `librist`, which includes important bug fixes and improvements.

### Motivation and Context
The previously used version of `librist` was outdated and caused runtime crashes (confirmed with some versions >= 0.2.9).  
This update points to a commit where those issues have been resolved.

See:  
https://code.videolan.org/rist/librist/-/commit/<commit_id>

### How Has This Been Tested?
- Successfully rebuilt FFmpeg with the updated `librist` hash on macOS Sequoia (ARM64).
- Rebuilt OBS with the updated FFmpeg.
- Tested RIST output in OBS; confirmed that the previous crash no longer occurs.
- No regressions or side effects observed in other features.

### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Code cleanup (non-breaking change that updates a dependency)
- [ ] Documentation (non-user-facing update to scripts)

### Checklist:
- [x] My code has been run through `clang-format`

### DIFF
--- a/deps.ffmpeg/70-librist.ps1
+++ b/deps.ffmpeg/70-librist.ps1
@@ -2,7 +2,7 @@ param(
     [string] $Name = 'librist',
     [string] $Version = '0.2.7',
     [string] $Uri = 'https://code.videolan.org/rist/librist.git',
-    [string] $Hash = "809390b3b75a259a704079d0fb4d8f1b5f7fa956",
+    [string] $Hash = "186bedbb6a2ec77f994689ccdde4a9bd88548d22",
     [array] $Targets = @('x64', 'arm64'),
     [array] $Patches = @(
         @{
--- a/deps.ffmpeg/70-librist.zsh
+++ b/deps.ffmpeg/70-librist.zsh
@@ -4,7 +4,7 @@ autoload -Uz log_debug log_error log_info log_status log_output
 local name='librist'
 local version='0.2.7'
 local url='https://code.videolan.org/rist/librist.git'
-local hash='809390b3b75a259a704079d0fb4d8f1b5f7fa956'
+local hash='186bedbb6a2ec77f994689ccdde4a9bd88548d22'
 local -a patches=(
   "macos ${0:a:h}/patches/librist/0001-generate-cross-compile-files-macos.patch \
     e14ae6f6565c9412c3f99f3917c1e8410181faede1530cbc014d2d5e03f9c124"


pkviet:
Did you test the deps building on the other platforms ? windows in particular.
The main blocker preventing the update has been a nasty windows bug which took ages to be pushed into release.
I think the version string should be updated, but i'm not sure how we deal with it when the hash is on master. I'll defer to @RytoEX on this.
We definitely prefer to ship a release version.

lescurej:
not tested on windows, but on mac it avoid OBS crashing after a long run when using librist in reception

deps.ffmpeg/70-librist.ps1
```diff
@@ -2,7 +2,7 @@ param(
     [string] $Name = 'librist',
     [string] $Version = '0.2.7',
```
RytoEX, line 3: Update version number.

deps.ffmpeg/70-librist.zsh
```diff
@@ -4,7 +4,7 @@ autoload -Uz log_debug log_error log_info log_status log_output
 local name='librist'
 local version='0.2.7'
```
RytoEX, line 5: Update version number.

RytoEX (changes requested):
https://github.com/obsproject/obs-deps/pull/228#issuecomment-1897611888
> These did not work:
>    Streaming
>        FAILS: RIST (local network)
> 
> I'm not sure about the RIST failure. It's possible that I misconfigured something, so additional testing would be appreciated. Receiving RIST seems to work fine though. cc @pkviet on this. If we can't resolve it quickly, I'll just revert the librist update.

https://github.com/obsproject/obs-deps/pull/245:
> I did not update librist because versions 0.2.8-0.2.10 have a bug, and there is no new tag. See:
> * https://github.com/obsproject/obs-deps/pull/228

@pkviet is correct that I prefer a stable tag, so v0.2.11 is what should be targeted and tested. This _must_ be tested on Windows, where I observed the previous failures, and it must be tested in both sending and receiving scenarios.

> See:
> https://code.videolan.org/rist/librist/-/commit/<commit_id>

Please actually specify the commit hash or tag in the PR description.

Please see commit history for the correct commit prefix/subject format.

RytoEX:
Closing this as there has been no response to feedback.

PR #303, opened 2025-11-04
Add Steinberg VST3 SDK

pkviet:
### Description
This adds Steinberg VST3 SDK to obs-deps in the include subfolder of the zips.

### Motivation and Context
Allows building of obs-vst3, cf : https://github.com/obsproject/obs-studio/pull/12752

### How Has This Been Tested?
Tested on my fork; successfully adds the SDK and allows building of obs-vst3 on windows and macOS.

### Types of changes

### Checklist:

### DIFF
--- a/deps.macos/90-vst3.zsh
+++ b/deps.macos/90-vst3.zsh
@@ -1,0 +1,33 @@
+autoload -Uz log_debug log_error log_info log_status log_output
+
+## Dependency Information
+local name='vst3sdk'
+local version='3.8.0'
+local url='https://github.com/steinbergmedia/vst3sdk.git'
+local hash="9fad9770f2ae8542ab1a548a68c1ad1ac690abe0"
+
+## Build Steps
+setup() {
+  log_info "Setup (%F{3}${target}%f)"
+  setup_dep ${url} ${hash}
+}
+
+install() {
+  log_info "Install (%F{3}${target}%f)"
+  cd ${dir}
+
+  mkdir -p \
+    ${target_config[output_dir]}/include \
+    ${target_config[output_dir]}/licenses/vst3sdk
+
+  mkdir -p ${target_config[output_dir]}/include/vst3sdk
+  cp -R base ${target_config[output_dir]}/include/vst3sdk/
+  cp -R pluginterfaces "${target_config[output_dir]}/include/vst3sdk/"
+
+  mkdir -p ${target_config[output_dir]}/include/vst3sdk/public.sdk
+  cp -R public.sdk/source ${target_config[output_dir]}/include/vst3sdk/public.sdk/
+
+  if [[ -f "LICENSE.txt" ]]; then
+    cp LICENSE.txt ${target_config[output_dir]}/licenses/vst3sdk/
+  fi
+}
--- a/deps.windows/90-vst3.ps1
+++ b/deps.windows/90-vst3.ps1
@@ -1,0 +1,61 @@
+param(
+    [string] $Name = 'vst3sdk',
+    [string] $Version = 'v3.8.0',
+    [string] $Uri = 'https://github.com/steinbergmedia/vst3sdk.git',
+    [string] $Hash = '9fad9770f2ae8542ab1a548a68c1ad1ac690abe0',
+    [array] $Targets = @('x64', 'arm64')
+)
+
+function Setup {
+    Log-Information "Setup (${Target})"
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
+}
+
+function Install {
+    Log-Information "Install (${Target})"
+    Set-Location $Path
+
+    $Params = @{
+        ErrorAction = "SilentlyContinue"
+        Path        = @(
+            "$($ConfigData.OutputPath)/include",
+            "$($ConfigData.OutputPath)/licenses"
+        )
+        ItemType    = "Directory"
+        Force       = $true
+    }
+    New-Item @Params *> $null
+
+    $Items = @(
+        @{
+            Path        = "base"
+            Destination = "$($ConfigData.OutputPath)/include/vst3sdk/base"
+            Recurse     = $true
+            ErrorAction = 'SilentlyContinue'
+        }
+        @{
+            Path        = "pluginterfaces"
+            Destination = "$($ConfigData.OutputPath)/include/vst3sdk/pluginterfaces"
+            Recurse     = $true
+            ErrorAction = 'SilentlyContinue'
+        }
+        @{
+            Path        = "public.sdk/source"
+            Destination = "$($ConfigData.OutputPath)/include/vst3sdk/public.sdk/source"
+            Recurse     = $true
+            ErrorAction = 'SilentlyContinue'
+        }
+        @{
+            Path        = "LICENSE.txt"
+            Destination = "$($ConfigData.OutputPath)/licenses/vst3sdk"
+            Force       = $true
+            ErrorAction = 'SilentlyContinue'
+        }
+    )
+
+    $Items | ForEach-Object {
+        $Item = $_
+        Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
+        Copy-Item @Item
+    }
+}


deps.macos/90-vst3.zsh
```diff
@@ -0,0 +1,42 @@
...
+  log_info "Setup (${target})"
+  if [[ -d "${dir}" ]]; then
+    rm -rf "${dir}"
+  fi
+  mkcd "${dir}"
+  git clone --filter=blob:none --sparse --recurse-submodules=no "${url}" .
+  git checkout "${hash}"
+  git submodule update --init base pluginterfaces public.sdk
+  rm -rf .git
+}
```
PatTheMav, line 13: Use `setup_dep` here, if the sparse checkout is strictly necessary you can use `dep_checkout` with the `--sparse` argument.

That way the checkout should always work and especially submodule changes would be covered automatically (e.g. if the submodules were to be replaced with in-tree sources, this code here would need to be changed, whereas using `dep_checkout` it would be guaranteed to work without changes).

deps.macos/90-vst3.zsh
```diff
@@ -0,0 +1,42 @@
...
+  mkdir -p "${target_config[output_dir]}/include/vst3sdk"
+  cp -R base "${target_config[output_dir]}/include/vst3sdk/"
+  cp -R pluginterfaces "${target_config[output_dir]}/include/vst3sdk/"
+
+  mkdir -p "${target_config[output_dir]}/include/vst3sdk/public.sdk"
+  cp -R public.sdk/source "${target_config[output_dir]}/include/vst3sdk/public.sdk/"
+
+  if [[ -f "LICENSE.txt" ]]; then
+    cp LICENSE.txt "${target_config[output_dir]}/licenses/vst3sdk/"
+  fi
```
PatTheMav, line 32: You can omit double quoting paths in Zsh as it won't treat spaces in variables as a field separators (in contrast to Bash). See also the other Zsh-based build scripts' file operations.

deps.windows/90-vst3.ps1
```diff
@@ -0,0 +1,76 @@
...
+        Remove-Item -Recurse -Force $Path
+    }
+    New-Item -Path $Path -ItemType Directory -Force *> $null
+    Set-Location $Path
+
+    Invoke-External git clone --filter=blob:none --sparse --recurse-submodules=no $Uri .
+    Invoke-External git checkout $Hash
+    Invoke-External git submodule update --init base pluginterfaces public.sdk
+
+    if (Test-Path ".git")        { Remove-Item ".git" -Recurse -Force }
```
PatTheMav: Same as with the Zsh variant, use `Invoke-GitCheckout` here (it also has the optional `-Sparse` argument, if it is necessary).

deps.windows/90-vst3.ps1
```diff
@@ -0,0 +1,76 @@
...
+        Force       = $true
+    }
+    New-Item @Params *> $null
+
+    $Items = @(
+        @{
+            Path        = "base"
+            Destination = "$($ConfigData.OutputPath)/include/vst3sdk/base"
+            Recurse     = $true
+            Container   = $true
```
PatTheMav: What is the purpose of explicitly enabling `Container` for each copy operation? It's probably the default anyway, but would there be a negative impact if it were omitted and `$false` became the default?

deps.windows/90-vst3.ps1
```diff
@@ -0,0 +1,76 @@
...
+        }
+    )
+
+    $Items | ForEach-Object {
+        $Item = $_
+        Log-Output ('{0} => {1}' -f ($Item.Path -join ", "), $Item.Destination)
+        Copy-Item @Item
+    }
+
+    Log-Information "$Name $Version installation complete."
```
PatTheMav: We don't use final completion messages in the build scripts, so this should be removed.

In contrast to the macOS build scripts, the Windows variant does not print a "Done" message itself, but that's an oversight that can be fixed in the framework separately then.

deps.macos/90-vst3.zsh
```diff
@@ -0,0 +1,42 @@
...
+
+  log_info "${name} ${version} installation complete."
```
PatTheMav: The framework will print a `DONE` message itself, using the package name as the message prefix, so this is duplicate output and can be removed.

deps.macos/90-vst3.zsh
pkviet: Does setup_dep work for pulling just some submodules but not all ?
I am not pulling the whole sdk, there's a lot of unnecessary stuff like docs, samples ...

PatTheMav: That shouldn't be a problem because the install step extracts the stuff needed for OBS-Studio anyway and such an optimisation is not necessary for `obs-deps` which runs at most once per day (and even then just to check if anything breaks) and even then has to pull much more data and submodules for stuff like `nlohmann-json` (because of its git history), `libdatachannel`, or even `qt`, already.

pkviet: fixed

deps.macos/90-vst3.zsh
pkviet: ah nice

deps.macos/90-vst3.zsh
pkviet: fixed

deps.windows/90-vst3.ps1
pkviet: fixed

deps.windows/90-vst3.ps1
pkviet: correct, it is already the default so no need to specify it indeed. Fixed

deps.windows/90-vst3.ps1
pkviet: fixed

deps.windows/90-vst3.ps1
```diff
@@ -0,0 +1,61 @@
...
+    [string] $Name = 'vst3sdk',
+    [string] $Version = 'v3.8.0',
+    [string] $Uri = 'https://github.com/steinbergmedia/vst3sdk.git',
+    [string] $Hash = '9fad9770f2ae8542ab1a548a68c1ad1ac690abe0',
+    [array] $Targets = @('x64', 'arm64')
+)
+
+function Setup {
+    Log-Information "Setup (${Target})"
+    Setup-Dependency -Uri $Uri -Hash $Hash -DestinationPath $Path
```
RytoEX, line 11: It _may_ be faster to manually call `Invoke-GitCheckout` with `-Sparse`:
https://github.com/obsproject/obs-deps/blob/2a08bc9f5996efb0627a66dacdb74727d8c79459/deps.ffmpeg/80-amf.ps1#L14

Otherwise, we'd have to update `Setup-Dependency` to accept the `-Sparse` parameter and arguments.

RytoEX (changes requested):

deps.windows/90-vst3.ps1
RytoEX: See also the macOS variant:
https://github.com/obsproject/obs-deps/blob/2a08bc9f5996efb0627a66dacdb74727d8c79459/deps.ffmpeg/80-amf.zsh#L16

pkviet: I initially used a sparse checkout, see the exchange with Pat. But he advised to use Setup-Dependency, arguing that speed doesn't matter. See: https://github.com/obsproject/obs-deps/pull/303#discussion_r2494096494 and following exchange

RytoEX: It seems what he said was to use `dep_checkout` or `Invoke-GitCheckout` if a sparse checkout is necessary, rather than manually writing `git` commands. I haven't tested a sparse checkout locally to see how it handles the submodules. I'll think about it.

pkviet: > It seems what he said was to use `dep_checkout` or `Invoke-GitCheckout` if a sparse checkout is necessary, rather than manually writing `git` commands. I haven't tested a sparse checkout locally to see how it handles the submodules. I'll think about it.

I quote:
`Use setup_dep here, if the sparse checkout is strictly necessary you can use dep_checkout with the --sparse argument.`

It seems it was ok to use either Setup-Dependency or dep_checkout  --sparse .
I went with the first; do you advise to use rather dep_checkout  --sparse ?

RytoEX: It's mostly about whether or not the time/disk savings are considered "necessary". I tend to prioritize those savings within reason, whereas Pat prefers less manual tweaks and less divergence overall particularly for things like obs-deps that run infrequently. In this case, I would accept either solution. We can always switch to `dep_checkout` and `Invoke-GitCheckout` or investigate the savings at a later time.

pkviet: gotcha, i've started to look into switching to Invoke-GitCheckout sparse but it doesn't work because the subfolders are actually pulled as submodules so the sparse option fails. Compare with my earlier manual call to git: 
>git clone --filter=blob:none --sparse --recurse-submodules=no "${url}" .     
git checkout "${hash}"     
git submodule update --init base pluginterfaces public.sdk

while Invoke-GitCheckout.ps1 has:
>if ( $Sparse ) {
            Invoke-External git clone --filter=blob:none --no-checkout $Uri $Path  
            Push-Location -Stack GitCheckoutTemp -Path $Path  
            Invoke-External git sparse-checkout @SparseArgs  
            Pop-Location -Stack GitCheckoutTemp  
        }  

dep_checkout behaves in the same manner.
It doesn't seem worth it to modify these scripts to accomodate the case of Steinberg sdk which is heavily submoduled.
So I shall stick to Setup-Dependency and setup-dep scripts.

RytoEX (approved):
Probably fine as-is. I will investigate sparse-checkout with submodules at a later date.

PR #309, opened 2026-04-06
utils.pwsh: Fix Windows config paths

RytoEX:
### Description

Apparently, CMake can fail when parsing the current path string because it interprets the backslash as the start of an escape sequence. Modern Windows can interpret paths with forward slashes, so it should be safe to change these slashes.

### Motivation and Context

Discovered this while attempting to update dependencies. Want to fix it separate from that update.

https://github.com/RytoEX/obs-deps/actions/runs/23963204885/job/69897580516#step:4:6211

### How Has This Been Tested?

I haven't tested this change yet, but we use forward slashes in other paths for Windows elsewhere, so this is hopefully fine.

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Setup-Host.ps1
+++ b/utils.pwsh/Setup-Host.ps1
@@ -20,7 +20,7 @@ function Setup-Host {
         $script:ProjectRoot = $($script:PSScriptRoot)
     }
 
-    $script:WorkRoot = "${ProjectRoot}\windows_build_temp"
+    $script:WorkRoot = "${ProjectRoot}/windows_build_temp"
 
     if ( ! ( $script:SkipAll ) && ( $script:SkipDeps ) ) {
         if ( ! ( Test-Path function:Install-BuildDependencies ) ) {
--- a/utils.pwsh/Setup-Target.ps1
+++ b/utils.pwsh/Setup-Target.ps1
@@ -27,7 +27,7 @@ function Setup-Target {
     $script:ConfigData = $TargetData[$script:Target]
 
     $script:ConfigData += @{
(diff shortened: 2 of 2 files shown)


PatTheMav (approved):
I'm almost of the opinion that within PowerShell-scripts as well as CMake code we should default to POSIX path separators as that makes paths identical across all platforms and cannot accidentally introduce parsing issues due to escaping or be interpreted as a command line argument on Windows, etc.

PatTheMav:
Possibly out of scope, but I also noticed that if `git` is installed within the `msys2` environment but also "Git for Windows" is installed, that any invocation of `git` might hit the former, but not the latter, and provided path strings are misinterpreted.

I ended up with several distributed path hierarchies starting with a root `c` directory on my dev machine, depending from which directory I invoked the `obs-deps` build script.

RytoEX:
> I'm almost of the opinion that within PowerShell-scripts as well as CMake code we should default to POSIX path separators as that makes paths identical across all platforms and cannot accidentally introduce parsing issues due to escaping or be interpreted as a command line argument on Windows, etc.

I agree, but that was a more sweeping change than I wanted to attempt at this time.


> Possibly out of scope, but I also noticed that if `git` is installed within the `msys2` environment but also "Git for Windows" is installed, that any invocation of `git` might hit the former, but not the latter, and provided path strings are misinterpreted.
> 
> I ended up with several distributed path hierarchies starting with a root `c` directory on my dev machine, depending from which directory I invoked the `obs-deps` build script.

Yes, I would say that's out of scope for this PR. I think I have previously noticed this, but it has been some time. I think it depends on the order of your PATH entries.

PR #310, opened 2026-04-06
utils.pwsh: Fix conditional expression

### LINKED ISSUE #307: Condtional in Setup-Host.ps1 always true
https://github.com/obsproject/obs-deps/blob/2a08bc9f5996efb0627a66dacdb74727d8c79459/utils.pwsh/Setup-Host.ps1#L25C6-L25C59

The above conditional uses the && pipeline chain operator, but I believe `-and` was meant to be used here. Currently this block of code will always execute, no matter the boolean value of `$SkipAll` and `$SkipDeps`.

RytoEX:
### Description

Fix a conditional expression in utils.pwsh/Setup-Host.ps1.

### Motivation and Context

Fixes #307.

### How Has This Been Tested?

Tested locally in PowerShell with simple if statements.
```pwsh
if ( $false && $false ) { echo "true" }
```
Outputs `true`.

```pwsh
if ( $false -and $false ) { echo "true" }
```
Does not output anything.

### Types of changes

### Checklist:

### DIFF
--- a/utils.pwsh/Setup-Host.ps1
+++ b/utils.pwsh/Setup-Host.ps1
@@ -22,12 +22,14 @@ function Setup-Host {
 
     $script:WorkRoot = "${ProjectRoot}\windows_build_temp"
 
-    if ( ! ( $script:SkipAll ) && ( $script:SkipDeps ) ) {
+    if ( ! ( $script:SkipAll -or $script:SkipDeps ) ) {
         if ( ! ( Test-Path function:Install-BuildDependencies ) ) {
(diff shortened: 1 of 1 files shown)


RytoEX:
After the initial CI failures, I noticed that this was a bit more nuanced than a simple replacement. The intent seems to be to skip `Install-Dependencies` if either `-SkipAll` or `-SkipDeps` is specified. Update the conditional logic to reflect that, and add debug logging to the `else` branch to verify.

PatTheMav (approved):

